PR #80, opened 2020-10-16
Update config key for legacy build

nporter-adbe:


emdobrin (commented):
Can you also update the event type to com.adobe.eventType.edge

nporter-adbe:
updated

emdobrin (approved):

PR #102, opened 2020-11-06
Add the support for the `query` parameter

shalehaha:
Add an optional field `query` to ExperienceEvent


emdobrin (commented):
Changes look good, left some small comments.

shalehaha:
@emdobrin fixed

emdobrin (approved):

Sources/EdgeNetworkHandlers/RequestBuilder.swift
```diff
@@ -97,6 +97,10 @@ class RequestBuilder {
                 eventData[Constants.JsonKeys.XDM] = xdm
             }
 
+            if let query = eventData[Constants.JsonKeys.QUERY] as? [String: Any] {
```
kevinlind, line 100: Let's remove this as it appears to just set the same value back to `eventData`. I'm seeing the tests pass after removing this statement.

kevinlind (approved):

Issue #96, opened 2020-11-05
Implement data persistence for xdm events


emdobrin:
Requirements:
- Use DataQueue service in the Edge extension to persist the xdm events received. 
- The persisted event should contain all the required information at the time it is batched as described here:
      - event data (identities, xdm and free form data) 
      - config id
      - dataset id (experience or profile)
      - Request event UUID
      - Event timestamp
- Events queueing should happen at all times as follows: when a new platform event is received by the ExperiencePlatform extension the event is queued persistently and then an attempt to send it to ExEdge server is made.
- The events are queued and processed (sent to ExEdge) in the same order they were received. If an event requires retry due to unavailable network connection or a recoverable error, then the other events wait in the queue until the first event is processed.
- The response callback will be invoked for the events processed in the same session when the request event was received.
- Other details: no batch limit will be used, retry for an event will happen until success/unrecoverable error, the events will not expire.

Retry definition:
- Recoverable error codes: 502, 503 (respect the Retry-After header), 504, 429 (should respect the Retry-After response header)
- Unrecoverable error codes: 500, 4xx
- 207 non-fatal error, wil just log the error/warning

Include unit tests as part of the feature implementation.

Part of https://github.com/adobe/aepsdk-edge-ios/issues/82

Issue #91, opened 2020-10-22
Update error network response handling


emdobrin:
Details:
- handle fatal errors, non-200
- 207 multi-status, at least one error in the body
- 200, may have warnings in the body

New error format:
```
"errors" : [
   {
      "type": "namespacedErrorCode",
      "status": 503,
      "title": "The service is temporarily unable to serve this request. Please try again later."
    }
  ]
```
New warning format:
```
"warnings": [
    {
      "type": "namespacedErrorCode",
      "status": 200,
      "title": "A warning occurred while calling the service.",
      "report": {
        "cause": {
          "message": "Cannot read related id...",
          "code": 202
        }
      }
    }
  ]
```

nporter-adbe:
Will pick this up and begin progress once https://github.com/adobe/aepsdk-edge-ios/pull/103 lands.

emdobrin:
Part of https://github.com/adobe/aepsdk-edge-ios/issues/125

Issue #90, opened 2020-10-21
Revisit EdgeResponseHandler


emdobrin:
As we start having more events which return event handles back to the client, we should evaluate if the EdgeResponseHandler is fulfilling the needs. Discuss about onError, onComplete callback.
Evaluate if a regular completion handler can be used instead of a protocol.

shalehaha:
Prefer using a closure instead of a protocol. 
Definitely need  onError, onComplete callback, or we can combine all the things into one closure, something like:
```
(data, error, isLast ) -> Void
```

Another thing I want to discuss, with today's API,`EdgeResponseHandler` is designed to be called multiple times. It is not easy for a client (customer or other extension built on top of edge) to handle the responses this way, it is possible for edge to wait for all the responses and then return them as an array to the caller?

shalehaha:
```
public protocol EdgeResponseHandler {
    func onPartialResponse(data: [String: Any])
    func onComplete()
    func onErrors(errors:[EdgeEventError])
}
```
@emdobrin according to our discussion before, is this the correct way how we want to change the `EdgeResponseHandler`?

Do we need to differentiate between errors vs warnings?

emdobrin:
@shalehaha here are the two options I think that cover all the use-cases you added in here:

1. Incremental updates, get advantage of the streaming option and get updated right away as we continue to process the data: 
```
public protocol EdgeResponseHandler {
    func onResponseUpdate(eventHandle: EdgeEventHandle) // may be called multiple times
    func onErrorUpdate(error: EdgeEventError) // may be called multiple times
    func onComplete() // called when the processing is over, onResponseUpdate and onErrorUpdate will not be called anymore after this
}
static func sendEvent(experienceEvent: ExperienceEvent, responseHandler: EdgeResponseHandler? = nil) // unchanged

Usage:
Edge.sendEvent(experienceEvent: experienceEvent, responseHandler: DemoResponseHandler())
```
2. Have a completion closure which returns all the handles / errors for the corresponding event as lists once the Edge extension finished processing the server response.
```
static func sendEvent(experienceEvent: ExperienceEvent, completion: (([EdgeEventHandle], [EdgeEventError]) -> Void)? = nil) {
    }

Usage:
Edge.sendEvent(experienceEvent: experienceEvent) { (handles: [EdgeEventHandle], errors: [EdgeEventError]) in
     //code
}
```
[...]

PR #103, opened 2020-11-09
Persist Edge Hits in Queue

nporter-adbe:
This PR adds support for persisting an `EdgeHit` to disk and processing each hit through an asynchronous queue. Each queue is provided a `HitProcessor`, and in this case an `EdgeHitProcessor`.  The `EdgeHitProcessor` is responsible for determining if we should move to the next hit, or if we should retry a given hit based on the `EdgeResponse`.


Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
```diff
@@ -389,6 +390,8 @@ class AEPEdgeFunctionalTests: FunctionalTestBase {
         XCTAssertEqual(7, requestBody.count)
         resetTestExpectations()
 
+        sleep(2)
```
nporter-adbe, line 393: This is a bit of an edge case in testing. This specific test case requires that we get a response for the first event and update the store payloads, then send out the second event with the store payloads. Everything was synonymous with the original implementation, so this worked fine; however, it is asynchronous with the HitQueue, so we need to add a sleep between the first and second events to give it a chance to process the hit and response.

kevinlind (approved):
Looks good to me.

Tests/UnitTests/EdgeHitProcessorTests.swift
```diff
@@ -0,0 +1,129 @@
...
+        // verify
+        wait(for: [expectation], timeout: 1)
+        XCTAssertTrue(mockNetworkService?.connectAsyncCalled ?? false) // network request should have been made
+        XCTAssertEqual(mockNetworkService?.connectAsyncCalledWithNetworkRequest?.url, expectedUrl) // network request should be made with the url in the hit
+    }
+
+    /// Tests that when the network request fails but has a recoverable error that we will retry the hit and do not invoke the response handler for that hit
+    func testProcessHitRecoverableNetworkError() {
+        // setup
+        let expectation = XCTestExpectation(description: "Callback should be invoked with true signaling this hit should be retried")
```
emdobrin, line 79: ..should be invoked with false...

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -0,0 +1,74 @@
...
+            return
+        }
+
+        callback.onError(jsonError: jsonError)
+        let isRecoverable = edgeErrorResponse.errors?.contains(where: {$0.isRecoverable}) ?? false
+        completion(!isRecoverable) // error, retry this hit if it is recoverable
+    }
+
+    func onComplete() {
+        callback.onComplete()
```
kevinlind, line 71: I was testing with this code and at some point I was getting an empty response back from Konductor. What I saw was only `onComplete` was called, but not `onResponse` or `onError` which meant the hits in the queue were not being removed. I would suggest adding `completion(true)` here as a fail-safe.

Tests/UnitTests/EdgeHitProcessorTests.swift
```diff
@@ -0,0 +1,129 @@
...
+            expectation.fulfill()
+        }
+
+        // verify
+        wait(for: [expectation], timeout: 1)
+        XCTAssertFalse(mockNetworkService?.connectAsyncCalled ?? true) // no network request should have been made
+    }
+
+    /// Tests that when a good hit is processed that a network request is made and the request returns 200
+    func testProcessHitHappy() {
```
emdobrin, line 51: testProcessHit_happy_sendsNetworkRequest_returnsTrue

kevinlind (changes requested):
Found an issue after some more testing.

Tests/UnitTests/EdgeHitProcessorTests.swift
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        // verify
+        wait(for: [expectation], timeout: 1)
+        XCTAssertTrue(mockNetworkService?.connectAsyncCalled ?? false) // network request should have been made
+        XCTAssertEqual(mockNetworkService?.connectAsyncCalledWithNetworkRequest?.url, expectedUrl) // network request should be made with the url in the hit
+    }
+
+    /// Tests that when the network request fails but has a recoverable error that we will retry the hit and do not invoke the response handler for that hit
+    func testProcessHitRecoverableNetworkError() {
```
emdobrin, line 77: testProcessHit_whenRecoverableNetworkError_sendsNetworkRequest_returnsFalse

Tests/UnitTests/EdgeHitProcessorTests.swift
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        // verify
+        wait(for: [expectation], timeout: 1)
+        XCTAssertTrue(mockNetworkService?.connectAsyncCalled ?? false) // network request should have been made
+        XCTAssertEqual(mockNetworkService?.connectAsyncCalledWithNetworkRequest?.url, expectedUrl) // network request should be made with the url in the hit
+    }
+
+    /// Tests that when the network request fails and does not have a recoverable response code that we invoke the response handler and do not retry the hit
+    func testProcessHitUnrecoverableNetworkError() {
```
emdobrin, line 106: testProcessHit_whenUnrecoverableNetworkError_sendsNetworkRequest_returnsTrue

Tests/UnitTests/EdgeHitProcessorTests.swift
```diff
@@ -0,0 +1,129 @@
...
+    func testProcessHitRecoverableNetworkError() {
+        // setup
+        let expectation = XCTestExpectation(description: "Callback should be invoked with true signaling this hit should be retried")
+        guard let expectedUrl = URL(string: "adobe.com") else { XCTFail("Could not create URL"); return }
+        let expectedEvent = Event(name: "Hit Event", type: EventType.identity, source: EventSource.requestIdentity, data: nil)
+        let expectedHeaders = ["testHeaderKey": "testHeaderVal"]
+        let expectedRequest = EdgeRequest(meta: nil, xdm: nil, events: nil)
+        let hit = EdgeHit(url: expectedUrl, request: expectedRequest, headers: expectedHeaders, event: expectedEvent)
+
+        let edgeResponse = EdgeResponse(requestId: "test-req-id", handle: nil, errors: [EdgeEventError(eventIndex: 0, message: "test-err", code: "502", namespace: nil)], warnings: nil)
```
emdobrin, line 86: we should test all recoverable error codes somehow, either in this test or testing isRecoverable

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
nporter-adbe: Is there any concern with invoking the callback more than once in the case that we do
onResponse() -> onComplete() that would invoke the callback twice, same with onError() -> onComplete().

As a side note, if we moved away from the protocol pattern this would simplify this logic and might make it a bit safer.
https://github.com/adobe/aepsdk-edge-ios/issues/90#issuecomment-715585562

Sources/Edge.swift
```diff
@@ -104,27 +107,32 @@ public class Edge: NSObject, Extension {
...
+            return
+        }
+
+        let edgeHit = EdgeHit(url: url, request: requestPayload, headers: requestHeaders, event: event)
+        guard let edgeHitData = try? JSONEncoder().encode(edgeHit) else {
+            Log.debug(label: LOG_TAG, "handleExperienceEventRequest - Failed to encode edge hit: '\(event.id.uuidString)'.")
+            return
         }
+
+        let entity = DataEntity(uniqueIdentifier: requestId, timestamp: event.timestamp, data: edgeHitData)
```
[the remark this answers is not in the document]
nporter-adbe: Why? The entity.uniqueIdentifier isn't related to the event, and in this implementation, we use the entity.uniqueIdnetifier to read the callback.
`let callback: ResponseCallback = NetworkResponseCallback(requestId: entity.uniqueIdentifier, responseHandler: networkResponseHandler)`

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -0,0 +1,74 @@
...
+    func processHit(entity: DataEntity, completion: @escaping (Bool) -> Void) {
+        guard let data = entity.data, let edgeHit = try? JSONDecoder().decode(EdgeHit.self, from: data) else {
+            // can't convert data to hit, unrecoverable error, move to next hit
+            Log.debug(label: LOG_TAG, "processHit - Failed to decode edge hit '\(entity.uniqueIdentifier)'.")
+            completion(true)
+            return
+        }
+
+        let callback: ResponseCallback = NetworkResponseCallback(requestId: entity.uniqueIdentifier, responseHandler: networkResponseHandler)
+        let hitCallback = EdgeHitResponseCallback(completion: completion, callback: callback)
```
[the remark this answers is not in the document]
nporter-adbe: I wish we could simplify it, but we can't (I haven't found a way). This is a direct product of using a protocol based callback pattern instead of a traditional callback pattern seen through other Swift APIs. The real fix here is to move away from the protocol based callbacks and use completion handlers.

Tests/UnitTests/EdgeHitProcessorTests.swift
nporter-adbe: Hmmm, we're using a different naming convention in our tests in Swift, but I'll update these to match edge.

Sources/EdgeNetworkHandlers/EdgeHit.swift
```diff
@@ -0,0 +1,29 @@
...
+/// Struct which represents an Identity hit
+struct EdgeHit: Codable {
+    /// URL to be requested for this Edge hit
+    let url: URL
+
+    /// The `EdgeRequest` for the corresponding hit
+    let request: EdgeRequest
+
+    /// Request headers for this `EdgeHit`
+    let headers: [String: String]
```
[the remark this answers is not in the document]
nporter-adbe: I'm concerned with computing the headers later, specifically because that shared state of assurance or something else could change from when this hit is queued vs when it is actually processed.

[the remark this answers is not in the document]
nporter-adbe: If there's no concern with computing the headers later, we can easily move this into the EdgeHitProcessor.

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
nporter-adbe: We could, however, that'd increase waits for all other tests that may not need the additional sleep time.

Tests/UnitTests/EdgeHitProcessorTests.swift
nporter-adbe: On the fence here, we're essentially testing the hardcoded array in the EdgeEventError, I think duplicating this test for all values in that array is a bit overkill. We currently don't have any tests for the EdgeEventError, even if we wrote tests for `isRecoverable` we'd just be asserting the list of error codes against another hard coded list that we need to maintain is correct, not sure there' much value gain there.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
kevinlind: Yes, you are right. It's my understanding that `completion()` will remove the top hit in the queue so it would be bad to call it multiple times for the same event. However, `onResponse` and `onError` may not get called if there are errors sending the request but they may get called multiple times when response streaming is enabled and when batching experience events. I believe `onComplete` is always called once, so we should call `completion(true)` only there. Also, `onComplete` should get called once after any retries are attempted, so you don't need to add any retry logic, just pass `true` to `completion`.

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
nporter-adbe: I am playing around with this and it doesn't appear to make a difference. I increased the timeout to something high like 20 seconds and it seems to just execute immediately. Going to look into this.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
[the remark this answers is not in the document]
emdobrin: As we discussed, you might need to refactor the doRequest logic a bit to cover the retries until success/unrecoverable error. Additionally when there is no network connection we need to wait for a longer period of time before retrying. 
This might also allow you to remove the double callback in here.

Sources/EdgeNetworkHandlers/EdgeHit.swift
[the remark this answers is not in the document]
emdobrin: Taking into consideration that these hits can be queued for longer period of time, you will want to send the attach the latest connected session. By the time the event is processed you may be either disconnected from Assurance or using a different session, so the headers needs to be updated.

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
emdobrin: Adding a wait time for the network request shouldn't affect this, it shouldn't wait for the entire timeout if the request is executed earlier.

Tests/UnitTests/EdgeHitProcessorTests.swift
emdobrin: I think we need to test that the retry logic is correctly executed for the recoverable error codes. If the code changes at any point, the tests should be a safeguard of that.

nporter-adbe: I updated this test to iterate over all the possible known recoverable error codes to cover this case.

Sources/EdgeNetworkHandlers/EdgeHit.swift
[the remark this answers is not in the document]
nporter-adbe: Moved the header generation into the hit processor.

Sources/EdgeNetworkHandlers/EdgeNetworkService.swift
```diff
@@ -122,52 +91,44 @@ class EdgeNetworkService {
...
-
+                    responseCallback.onComplete()
+                    completion(true) // successful request, return true
                 } else if responseCode == HttpResponseCodes.noContent.rawValue {
                     // Successful collect requests do not return content
                     Log.debug(label: self.LOG_TAG, "doRequest - Collect connection to Experience Edge was successful.")
-
+                    responseCallback.onComplete()
+                    completion(true) // successful request, return true
                 } else if self.recoverableNetworkErrorCodes.contains(responseCode) {
```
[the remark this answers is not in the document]
nporter-adbe: I've updated to account for 207 in the recoverable network codes list. I'll update the reading errors/warnings in https://github.com/adobe/aepsdk-edge-ios/issues/91.

Sources/Edge.swift
```diff
@@ -94,37 +97,27 @@ public class Edge: NSObject, Extension {
...
-                                     requestHeaders: requestHeaders,
-                                     responseCallback: callback,
-                                     retryTimes: Constants.Defaults.NETWORK_REQUEST_MAX_RETRIES)
+        guard let requestPayload = requestBuilder.getRequestPayload(listOfEvents) else {
+            Log.debug(label: LOG_TAG,
+                      "handleExperienceEventRequest - Failed to build the request payload, dropping current event '\(event.id.uuidString)'.")
+            return
         }
+
+        let edgeHit = EdgeHit(configId: configId, requestId: UUID().uuidString, request: requestPayload, event: event)
```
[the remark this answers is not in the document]
nporter-adbe: I've opted to build the request first as it is much safer than persisting the `Event`/list of events and decoding them later. The reason for this is that it could be possible that some of the event/xdm data is not compliant with `AnyCodable`, in the case that the event has data that can not be encoded/decoded with `AnyCodable`, that data is thrown out during encoding/decoding. I'd prefer to migrate the hits afterward then to hope that all the data is compliant with `AnyCodable` to minimize the risk of losing data.

For reference, the types that `AnyCodable` can handle can be found here:
https://github.com/adobe/aepsdk-core-ios/blob/main/AEPServices/Sources/utility/AnyCodable.swift#L84

Sources/Edge.swift
```diff
@@ -94,37 +97,27 @@ public class Edge: NSObject, Extension {
...
+            return
+        }
+
+        // NOTE: the order of these events needs to be maintained as they were sent in the network request
+        // otherwise the response callback cannot be matched
+        networkResponseHandler.addWaitingEvents(requestId: edgeHit.requestId,
+                                                batchedEvents: listOfEvents)
+
+        let entity = DataEntity(uniqueIdentifier: event.id.uuidString, timestamp: event.timestamp, data: edgeHitData)
+        hitQueue?.queue(entity: entity)
```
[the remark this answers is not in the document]
nporter-adbe: I think in the future we could modify the PersistentHitQueue to have a batch limit and only start processing after n hits are queued or something.

Sources/Edge.swift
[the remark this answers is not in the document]
nporter-adbe: Actually, this shouldn't make much of a difference as `EdgeRequest` contains `let events: [[String: AnyCodable]]?`. I'll update the PR.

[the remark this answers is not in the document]
nporter-adbe: Do we think it's possible to have data that falls outside the bounds of AnyCodable?

nporter-adbe:
@emdobrin thanks for the feedback, moved all the hit building into the hit processor.

Sources/EdgeNetworkHandlers/EdgeResponse.swift
```diff
@@ -14,7 +14,7 @@ import Foundation
 
 /// A response from the Adobe Experience Edge server.
 /// An `EdgeResponse` is the top-level response object received from the server.
-struct EdgeResponse: Decodable {
+struct EdgeResponse: Codable {
```
nporter-adbe, line 17: Only Codable for testing purposes.

Sources/EdgeNetworkHandlers/EdgeNetworkService.swift
```diff
@@ -22,18 +22,14 @@ enum ExperienceEdgeRequestType: String {
...
-enum RetryNetworkRequest: Int {
-    case yes
-    case no
-}
-
 /// Convenience enum for the known error codes
 enum HttpResponseCodes: Int {
     case ok = 200
     case noContent = 204
+    case multiStatus = 207
```
nporter-adbe, line 29: Will change the handling for 207 in another PR.

nporter-adbe:
@emdobrin @kevinlind ready for a final look

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -0,0 +1,134 @@
...
+import AEPServices
+import Foundation
+
+/// A `HitProcessing` which handles the processing of `EdgeHit`s
+class EdgeHitProcessor: HitProcessing {
+    private let LOG_TAG = "EdgeHitProcessor"
+    private var networkService: EdgeNetworkService
+    private var networkResponseHandler: NetworkResponseHandler
+    private var getSharedState: (String, Event?) -> SharedStateResult?
+    let retryInterval = Constants.NetworkKeys.RETRY_INTERVAL
```
kevinlind, line 23: I don't see that this variable is used anywhere. I also don't see where the RETRY_INTERVAL is used except here. If the hit is retried, where do we check the number of retires?

nporter-adbe: You're right it's not directly used anywhere in our implementation of the hit processor, but it is accessed in the PersistHitQueue. It is a required property defined in `HitProcesesing`
https://github.com/adobe/aepsdk-core-ios/blob/main/AEPServices/Sources/utility/hitprocessor/HitProcessing.swift#L17

Then used here: https://github.com/adobe/aepsdk-core-ios/blob/main/AEPServices/Sources/utility/hitprocessor/PersistentHitQueue.swift#L73

nporter-adbe: As far as retries, it retries indefinitely based on the spec/ticket. https://github.com/adobe/aepsdk-edge-ios/issues/96

> Other details: no batch limit will be used, retry for an event will happen until success/unrecoverable error, the events will not expire.

kevinlind: 👍  Thanks!

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -0,0 +1,134 @@
...
+                                               lineFeed: Constants.Defaults.LINE_FEED)
+
+        // get ECID from Identity shared state, this should be resolved based on readyForEvent check
+        guard let identityState =
+                getSharedState(Constants.SharedState.Identity.STATE_OWNER_NAME,
+                               event)?.value else {
+            Log.warning(label: LOG_TAG,
+                        "processHit - Unable to process the event '\(event.id.uuidString)', " +
+                            "Identity shared state is nil.")
+            completion(true)
```
kevinlind, line 58: Is it possible for a hit to be processed before the Identity extension is fully initialized?  As the hit queue is setup and started from the Edge.init() ctor, I'm wondering, if an app is launched with persisted hits, could they get processed before Identity was able to save its first state? Since Edge is configured to handle events only when Configuration and Identity have shared states, maybe the same should be true here?

nporter-adbe: Yeah, this is a good catch, I see a few options:

1. Only start hit processing once `readyForEvent` would return true to indicate that all the required shared states are published.
2. Change line 58 in EdgeHitProcessor to `completion(false)` so that this hit will be retried until Identity has provided a shared state.

I'm leaning towards option 2. Thoughts?

Sources/Edge.swift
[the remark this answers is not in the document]
kevinlind: The allowed AnyCodable types should be sufficient, but there's always the chance a customer could pass in something else. I think we should document the allowed types in ExperienceEvent for cases where customers are building their own dictionaries or implementing their own Schema classes. I think we have a similar issue in V5 when serializing EventData to Variant types. 
Here, https://github.com/adobe/aepsdk-edge-ios/blob/dev/Sources/EdgeNetworkHandlers/RequestBuilder.swift#L111 If the EventData doesn't conform to AnyCodable, we drop the event.

kevinlind (approved):
Looks good. Thanks for all the revisions @nporter-adbe. To support batched events we'll need to make changes to the hit queue and processor but that should be done in another PR.

nporter-adbe:
@emdobrin let me know if you have any additional comments, plan to merge soon to continue with the error response handling updates.

Tests/FunctionalTests/util/FunctionalTestBase.swift
```diff
@@ -263,7 +263,7 @@ class FunctionalTestBase: XCTestCase {
         }
 
         for expectedRequest in expectedNetworkRequests {
-            let waitResult = expectedRequest.value.await(timeout: 2)
+            let waitResult = expectedRequest.value.await(timeout: 10)
```
emdobrin, line 266: Can we make the timeout configurable on this API if needed just for certain tests?

Sources/EdgeNetworkHandlers/EdgeHit.swift
```diff
@@ -0,0 +1,29 @@
...
+    let configId: String
+
+    /// Unique identifier for this hit
+    let requestId: String
+
+    /// The `EdgeRequest` for the corresponding hit
+    let request: EdgeRequest
+
+    /// The `Event` associated with the `EdgeHit`
+    let event: Event
```
emdobrin, line 28: Is this event used for anything else other than logging on fail here https://github.com/adobe/aepsdk-edge-ios/pull/103/files#diff-1f8b1d6008249e9d5807a9cdd4264cbcb5fdaee71d22df628697818ddaeeb39cR111-R117? 
I am thinking the Edge hit will contain a list of events in the future, but I wonder if this is required in the EdgeHit at all, we should be able to tie the responses back based on the requestId and the queued waiting events. If not required, let's remove it from here.

emdobrin (commented):
Changes look good with few comments.

Tests/FunctionalTests/util/FunctionalTestBase.swift
nporter-adbe: We could make it configurable, but all network requests in Edge are going to need an increased timeout due to the async nature of the hit queue.

Issue #108, opened 2020-11-13
Set eventSource to the EdgeEventHandle type in AEP Response Event Handle


emdobrin:
Send the Edge response handles as Events through the Event Hub using eventType : com.adobe.eventType.edge and eventSource: EdgeEventHandle.type (e.g. `personalization:decisions`).
This way an extension can listen for a particular response handle they are interested in.

Update existing and add new tests for the above.

PR #109, opened 2020-11-18
Set eventSource to the EdgeEventHandle type in AEP Response Event Handle

nporter-adbe:
Sets the event source for edge response events to the value of the handle type.
More info: https://github.com/adobe/aepsdk-edge-ios/issues/108

### DIFF
--- a/Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
+++ b/Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
@@ -212,7 +212,10 @@ class NetworkResponseHandler {
     private func dispatchResponseEventWithData(_ eventData: [String: Any], requestId: String, isErrorResponseEvent: Bool, eventSource: String?) {
         guard !eventData.isEmpty else { return }
         var source = isErrorResponseEvent ? Constants.EventSource.ERROR_RESPONSE_CONTENT : Constants.EventSource.RESPONSE_CONTENT
-        source = eventSource ?? source
+        if let eventSource = eventSource, !eventSource.isEmpty {
+            source = eventSource
+        }
+
         let responseEvent = Event(name: isErrorResponseEvent ? Constants.EventName.ERROR_RESPONSE_CONTENT : Constants.EventName.RESPONSE_CONTENT,
                                   type: Constants.EventType.EDGE,
                                   source: source,
--- a/Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
+++ b/Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
@@ -138,7 +138,7 @@ class NetworkResponseHandler {
             handleStoreEventHandle(handle: eventHandle)
 
             guard let eventHandleAsDictionary = try? eventHandle.asDictionary() else { continue }
-            dispatchResponseEvent(handleAsDictionary: eventHandleAsDictionary, requestId: requestId, requestEventId: requestEventId, eventType: eventHandle.type)
+            dispatchResponseEvent(handleAsDictionary: eventHandleAsDictionary, requestId: requestId, requestEventId: requestEventId, eventSource: eventHandle.type)
             ResponseCallbackHandler.shared.invokeResponseHandler(eventData: eventHandleAsDictionary, requestEventId: requestEventId)
         }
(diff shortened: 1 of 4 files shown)


kevinlind:
@nporter-adbe Can you investigate why the coverage dropped for this PR?  It appears all your changes have 100% coverage, but I wonder if the changes to the tests are causing a code path to no longer get executed.

nporter-adbe:
@kevinlind Yeah, that's interesting given the additions have 100% diff coverage. From what I can tell, it's caused by the fact I removed/simplified some code covered by testing, thus reducing the overall file size. Reducing the file size by removing tested code has resulted in the uncovered code to represent a larger portion of the file, which is NetworkResponseHandler.swift in this case.

kevinlind:
@nporter-adbe Okay, that makes sense. It's odd that Codecov is reporting `-17` lines because that appears to include lines removed and lines changed. The PR only removes about 6 lines.

shalehaha:
I'm still debating whether it should be set to event source or type.

Tests/FunctionalTests/NetworkResponseHandlerFunctionalTests.swift
```diff
@@ -277,7 +277,7 @@ class NetworkResponseHandlerFunctionalTests: FunctionalTestBase {
             "    }"
         networkResponseHandler.processResponseOnSuccess(jsonResponse: jsonResponse, requestId: "123")
 
-        let dispatchEvents = getDispatchedEventsWith(type: FunctionalTestConst.EventType.EDGE, source: FunctionalTestConst.EventSource.RESPONSE_CONTENT)
+        let dispatchEvents = getDispatchedEventsWith(type: FunctionalTestConst.EventType.EDGE, source: "state:store")
```
kevinlind, line 280: Can you include a few new tests with 1) handle type empty and 2) no handle type, and verify an event with source `FunctionalTestConst.EventSource.RESPONSE_CONTENT` is dispatched?

nporter-adbe:
@shalehaha My take is that `type` has been used to signify which extension the event is coming from, even though this is a bit backward from how I'd expect source and type to be populated. So in this case I think we went with keeping that standard by having type set to `com.adobe.edge`.

Tests/FunctionalTests/NetworkResponseHandlerFunctionalTests.swift
nporter-adbe: Added.

nporter-adbe:
@shalehaha Yeah, I can see both sides of the coin here, I wish we could revisit the definition of type/source for all events. My preference here is to keep as much consistency as possible and have `com.adobe.edge` be the type. It seems to not make much of a difference either way as long as we document them properly given that both options have valid reasons.

emdobrin:
@shalehaha we discussed this originally for the Edge extension, we wanted to swap the type and source as it feels like it makes more sense that way, but we gave up to that idea for consistency across extensions. I think it shouldn't matter that much since we are using constants anyways when registering the listeners.

kevinlind (approved):

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -137,8 +137,8 @@ class NetworkResponseHandler {
             let requestEventId = extractRequestEventId(forEventIndex: eventHandle.eventIndex, requestId: requestId)
             handleStoreEventHandle(handle: eventHandle)
 
-            guard let eventHandleAsDictionary = try? eventHandle.asDictionary() else { return }
-            dispatchResponseEvent(handleAsDictionary: eventHandleAsDictionary, requestId: requestId, requestEventId: requestEventId)
+            guard let eventHandleAsDictionary = try? eventHandle.asDictionary() else { continue }
+            dispatchResponseEvent(handleAsDictionary: eventHandleAsDictionary, requestId: requestId, requestEventId: requestEventId, eventSource: eventHandle.type)
```
emdobrin, line 141: Nit: can we add a comment in here explaining that the handle type is used as event source for response event filtering and it is expected to have a standardized format.

emdobrin (approved):
Looks good @nporter-adbe

Issue #95, opened 2020-11-05
Write functional tests with response callback


emdobrin:
Test response callback is called for multiple streams
Test sendEvent works with response callback / with null callback

Proposed test workflow:

- set valid configs
- mock network response - multiple chuncks as a response for event1
- send xdm event1
- set invalid config (pending/nil)
- send xdm event2
- check callback is invoked correcly for event1, check xdm event2 not processed

Resolve the todo for this task in NoConfigFunctionalTests.

Part of https://github.com/adobe/aepsdk-edge-ios/issues/88

PR #110, opened 2020-11-18
Update Edge responses support

emdobrin:
## Description

- Removed usage of ResponseHandler
- Updated public API  to include optional completion handler `sendEvent(experienceEvent: ExperienceEvent, _ completion: (([EdgeEventHandle]) -> Void)? = nil)`
- Made EdgeEventHandle public to be used in the completion handler
- Test updates, new unit tests
- More context https://github.com/adobe/aepsdk-edge-ios/issues/90

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


Sources/Edge+PublicAPI.swift
```diff
@@ -39,4 +39,24 @@ public extension Edge {
...
+            Log.debug(label: LOG_TAG, "Failed to dispatch the experience event because the XDM data was nil/empty.")
+            return
+        }
+
+        let event = Event(name: "AEP Request Event",
+                          type: Constants.EventType.EDGE,
+                          source: Constants.EventSource.REQUEST_CONTENT,
+                          data: eventData)
+
+        ResponseCallbackHandler.shared.registerCompletionHandler(requestEventId: event.id.uuidString, completionHandler: completion)
```
shalehaha, line 59: I think it would be easier if we use the EdgeResponseHandler here. Just need to define a EdgeResponseHandler which caches all the responses and error, and then call the `completion` on complete.

Sources/EdgeNetworkHandlers/ResponseCallbackHandler.swift
```diff
@@ -39,32 +51,93 @@ class ResponseCallbackHandler {
...
-    func invokeResponseHandler(eventData: [String: Any], requestEventId: String?) {
+    /// Updates the list of `EdgeEventError`(s) for current `requestEventId` and calls onErrorUpdate if a `ResponseHandler` is registered.
+    /// - Parameters:
+    ///   - eventError: newly received error/warning message
+    ///   - requestEventId: the request event identifier associated with this event handle
+    func eventErrorReceived(_ eventError: EdgeEventError, requestEventId: String?) {
         guard let unwrappedRequestEventId = requestEventId, !unwrappedRequestEventId.isEmpty else {
             return
         }
+        if edgeEventErrors[unwrappedRequestEventId] != nil {
```
nporter-adbe, line 114: Same here on the `if let` comment.

Tests/UnitTests/CompletionHandlerTests.swift
```diff
@@ -0,0 +1,190 @@
...
+        return try! JSONDecoder().decode(EdgeEventError.self, from: jsonData)
+    }()
+    // swiftlint:enable force_try
+    private let uniqueEventId = "123"
+
+    override func setUp() {
+        continueAfterFailure = false // fail so nil checks stop execution
+    }
+
+    func testRegisterCompletionHandler_thenEventHandlerReceived_completionCalled() {
```
nporter-adbe, line 37: In these tests, we should also set `assertForOverfullfill` to ensure they are only invoked once.
https://developer.apple.com/documentation/xctest/xctestexpectation/2806575-assertforoverfulfill

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -188,7 +188,7 @@ class NetworkResponseHandler {
         Log.trace(label: LOG_TAG, "dispatchEventErrors - Processing \(unwrappedErrors.count) errors(s) for request id: \(requestId)")
         for error in unwrappedErrors {
 
-            if let errorAsDictionary = try? error.asDictionary() {
+            if let errorAsDictionary = error.asDictionary() {
```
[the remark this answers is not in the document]
kevinlind: For me, Xcode is flagging this line as not needing the `try` statement, probably because it's using `AnyCodeable.asDictionary` though I would expect `EdgeEventEncodable.asDictionary` to be used. 
I think we should update `EdgeEventEncodable.asDictionary` to remove the throws clause and just return `nil` on error, same as `AnyCodable.asDictionary`.

Sources/EdgeNetworkHandlers/ResponseCallbackHandler.swift
```diff
@@ -39,32 +51,93 @@ class ResponseCallbackHandler {
...
+        Log.trace(label: TAG, "Removing completion handlers for Edge response with request unique id \(requestEventId).")
+    }
+
+    /// Updates the list of `EdgeEventHandle`(s) for current `requestEventId` and calls onResponseUpdate if a `ResponseHandler` is registered.
+    /// - Parameters:
+    ///   - eventHandle: newly received event handle
+    ///   - requestEventId: the request event identifier associated with this event handle
+    func eventHandleReceived(_ eventHandle: EdgeEventHandle, requestEventId: String?) {
+        guard let unwrappedRequestEventId = requestEventId, !unwrappedRequestEventId.isEmpty else { return }
+        if edgeEventHandles[unwrappedRequestEventId] != nil {
```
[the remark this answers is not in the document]
emdobrin: this would create a copy and in this case I need to mutate the original array

Sources/Edge+PublicAPI.swift
emdobrin: I kept the two separated in case we want to stick with one or the other. If we keep the completion handler then we can remove EdgeResponseHandler altogether.

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
```diff
@@ -26,18 +26,26 @@ class AEPEdgeFunctionalTests: FunctionalTestBase {
...
 
-        func onResponse(data: [String: Any]) {
-            onResponseReceivedData = data
+        func onResponseUpdate(eventHandle: EdgeEventHandle) {
+            onResponseEdgeHandle = eventHandle
             countDownLatch.countDown()
         }
 
+        func onErrorUpdate(error: EdgeEventError) {
+            // todo
```
nporter-adbe, line 42: Can we log tickets/issues for these?

Tests/UnitTests/MockEdgeResponseHandler.swift
```diff
@@ -16,9 +16,28 @@ import Foundation
...
+        onErrorUpdateData = error
+    }
+
+    func onComplete() {
+        onCompleteCalledTimes += 1
+    }
+}
+
+extension EdgeEventHandle {
+    func toDictionary() -> [String: Any]? {
```
nporter-adbe, line 39: I've seen this implemented a few times. Any way we can have a single source of truth for this function?

Tests/UnitTests/ResponseHandlerTests.swift
```diff
@@ -0,0 +1,199 @@
...
+
+import AEPCore
+@testable import AEPEdge
+import XCTest
+
+class ResponseHandlerTests: XCTestCase {
+    private let requestEventId = "requestEventId"
+    // swiftlint:disable force_try
+    private var eventHandle: EdgeEventHandle = {
+        let json: [String: Any] = ["payload": [["key1": "value1", "key2": 2]], "type": "testType"]
```
nporter-adbe, line 21: Let's use real-world values in the event handle and event error.

Sources/EdgeNetworkHandlers/ResponseCallbackHandler.swift
```diff
@@ -39,32 +51,93 @@ class ResponseCallbackHandler {
...
+
+        if let completionHandler = completionHandlers[requestEventId] {
+            completionHandler(edgeEventHandles[requestEventId] ?? [], edgeEventErrors[requestEventId] ?? [])
+            completionHandlers[requestEventId] = nil
+        }
+
+        edgeEventHandles[requestEventId] = nil
+        edgeEventErrors[requestEventId] = nil
+
+        Log.trace(label: TAG, "Removing completion handlers for Edge response with request unique id \(requestEventId).")
```
kevinlind, line 89: nit: As this is for either completion handlers or response handlers, maybe reword to "Removing callback handlers for Edge response ..." ?

Sources/EdgeNetworkHandlers/ResponseCallbackHandler.swift
```diff
@@ -39,32 +51,93 @@ class ResponseCallbackHandler {
...
-            Log.trace(label: TAG, "Removing callback for Edge response with request unique id \(requestEventId).")
         }
+
+        if let completionHandler = completionHandlers[requestEventId] {
+            completionHandler(edgeEventHandles[requestEventId] ?? [], edgeEventErrors[requestEventId] ?? [])
+            completionHandlers[requestEventId] = nil
+        }
+
+        edgeEventHandles[requestEventId] = nil
+        edgeEventErrors[requestEventId] = nil
```
kevinlind, line 87: When clearing values from the ThreadSafeDictionary objects, can you use `removeValue` instead of setting it to `nil`? My thinking is setting the value to `nil` will keep the element in the dictionary and not free up memory.

Sources/EdgeNetworkHandlers/ResponseCallbackHandler.swift
```diff
@@ -39,32 +51,93 @@ class ResponseCallbackHandler {
...
+    ///   - eventHandle: newly received event handle
+    ///   - requestEventId: the request event identifier associated with this event handle
+    func eventHandleReceived(_ eventHandle: EdgeEventHandle, requestEventId: String?) {
+        guard let unwrappedRequestEventId = requestEventId, !unwrappedRequestEventId.isEmpty else { return }
+        if edgeEventHandles[unwrappedRequestEventId] != nil {
+            edgeEventHandles[unwrappedRequestEventId]?.append(eventHandle)
+        } else {
+            edgeEventHandles[unwrappedRequestEventId] = [eventHandle]
+        }
+        invokeResponseHandler(eventHandle: eventHandle, eventError: nil, requestEventId: unwrappedRequestEventId)
```
kevinlind, line 103: If this requestEventId is keyed to a completionHandler, and not a responseHandler, then calling `invokeResponseHandler` will log that the response handler was not found, even though event handler was correctly processed. Instead, can another conditional statement be added to check if the requestEventId is in the responseHandlers dictionary, and if so just call `invokeResponseHandler` otherwise add the eventHandle to edgeEventHandles dictionary?

Sources/EdgeNetworkHandlers/ResponseCallbackHandler.swift
```diff
@@ -39,32 +51,93 @@ class ResponseCallbackHandler {
...
+    func eventErrorReceived(_ eventError: EdgeEventError, requestEventId: String?) {
         guard let unwrappedRequestEventId = requestEventId, !unwrappedRequestEventId.isEmpty else {
             return
         }
+        if edgeEventErrors[unwrappedRequestEventId] != nil {
+            edgeEventErrors[unwrappedRequestEventId]?.append(eventError)
+        } else {
+            edgeEventErrors[unwrappedRequestEventId] = [eventError]
+        }
+        invokeResponseHandler(eventHandle: nil, eventError: eventError, requestEventId: unwrappedRequestEventId)
```
kevinlind, line 119: Same comment as about regarding only calling `invokeResponseHandler` if the requestEventId is in the responseHandlers' dictionary.

Sources/EdgeNetworkHandlers/ResponseCallbackHandler.swift
emdobrin: ThreadSafeDictionary is using a Dictionary under the hood, so setting the value to nil is removing the key-value pair from the dictionary, as stated in the swift doc https://developer.apple.com/documentation/swift/dictionary. You can find more discussions on this thread https://forums.swift.org/t/dictionary-with-optional-values/2675/15.
I updated the implementation though so that it is independent on the internal implementation of ThreadSafeDictionary.

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
emdobrin: fixed

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -151,8 +150,12 @@ class NetworkResponseHandler {
...
     /// - Returns: the request event unique identifier for which this event handle was received, nil if not found
     private func extractRequestEventId(forEventIndex: Int?, requestId: String) -> String? {
-        guard let requestEventIdsList = getWaitingEvents(requestId: requestId),
-              let index = forEventIndex, index >= 0, index < requestEventIdsList.count else {
+        guard let requestEventIdsList = getWaitingEvents(requestId: requestId) else { return nil }
+
+        // Note: ExEdge does not return eventIndex when there is only one event in the request.
+        // The event handles and errrors are associated to that request event, so defaulting to 0 here.
+        let index = forEventIndex ?? 0
+        guard index >= 0, index < requestEventIdsList.count else {
```
emdobrin, line 158: Fixed 🐛  where the response was not returned in the callback when eventIndex was missing in the response.

emdobrin:
I pushed a few more tests to support these changes and the related https://github.com/adobe/aepsdk-edge-ios/issues/95

Tests/FunctionalTests/AEPEdgeResponseHandlerFunctionalTests.swift
```diff
@@ -0,0 +1,192 @@
...
+        let httpConnection: HttpConnection = HttpConnection(data: responseBodyWithHandleAndError.data(using: .utf8),
+                                                            response: HTTPURLResponse(url: edgeUrl,
+                                                                                      statusCode: 200,
+                                                                                      httpVersion: nil,
+                                                                                      headerFields: nil),
+                                                            error: nil)
+        let responseHandler = MockResponseHandler(expectedResponses: 1, expectedErrors: 1)
```
kevinlind, line 175: Here you are expecting 2 error handles, however you initialize the MockResponseHandler with `expectedErrors: 1`.  Should the expected errors be 2?

Tests/FunctionalTests/AEPEdgeResponseHandlerFunctionalTests.swift
```diff
@@ -0,0 +1,192 @@
...
+        // verify
+        assertNetworkRequestsCount()
+        responseHandler2.await()
+
+        XCTAssertEqual(1, responseHandler1.onResponseHandles.count)
+        XCTAssertEqual(0, responseHandler1.onErrorHandles.count)
+        XCTAssertTrue(responseHandler1.onCompleteCalled)
+        XCTAssertEqual(0, responseHandler2.onResponseHandles.count)
+        XCTAssertEqual(2, responseHandler2.onErrorHandles.count)
+        XCTAssertTrue(responseHandler2.onCompleteCalled)
```
kevinlind, line 164: Are the response handlers reversed here?  You initialize responseHandler1 with 2 expected errors but assert that there is 1 response handler and 0 error handlers. Similarly, responseHandler2 is initialized with 1 expected response handler but you assert it has 2 error handlers and 0 response handlers.

Sources/EdgeEventHandle.swift
```diff
@@ -14,14 +14,15 @@ import AEPServices
...
     /// Encodes the event to which this handle is attached as the index in the events array in EdgeRequest
     let eventIndex: Int?
 
     /// Payload type
-    let type: String?
+    public let type: String?
 
     /// Event payload
-    let payload: [[String: AnyCodable]]?
+    public let payload: [[String: AnyCodable]]?
```
nporter-adbe, line 27: If we're going to make this a public class I would advise us not to expose `AnyCodable`. The `AnyCodable` type should only be used mainly for internal encoding/decoding. 

Let's update this to be more similar to the `Event` struct in how it handles `AnyCodable`
https://github.com/adobe/aepsdk-core-ios/blob/main/AEPCore/Sources/eventhub/Event.swift#L32
https://github.com/adobe/aepsdk-core-ios/blob/main/AEPCore/Sources/eventhub/Event.swift#L111
https://github.com/adobe/aepsdk-core-ios/blob/main/AEPCore/Sources/eventhub/Event.swift#L124

emdobrin:
Based on our discussion we decided to:
- only keep one public API with completion handler
- not return the error information in the completion handler, it will still be dispatched as an event through event hub
@nporter-adbe @kevinlind please take a look at the updated code.

Sources/EdgeNetworkHandlers/CompletionHandlersManager.swift
```diff
@@ -0,0 +1,75 @@
...
+// OF ANY KIND, either express or implied. See the License for the specific language
+// governing permissions and limitations under the License.
+//
+
+import AEPCore
+import AEPServices
+import Foundation
+
+/// When registering a completion handler for a request event, all the event handles will be returned at once when the entire streamed response was processed by the AEP Edge extension.
+/// This class uses a `ThreadSafeDictionary` for the internal mapping.
```
nporter-adbe, line 18: nit: I don't think we need to add docs regarding the internal implementation of the class.

Sources/EdgeNetworkHandlers/CompletionHandlersManager.swift
```diff
@@ -0,0 +1,75 @@
...
+    static let shared = CompletionHandlersManager()
+
+    /// Registers a completion handler for the specified `requestEventId`. This handler is invoked when the Edge response content has been
+    /// handled entirely by the Edge extension, containing a list of `EdgeEventHandle`(s). This list can be empty or can contain one or multiple items
+    /// based on the request and the server side response.
+    ///
+    /// - Parameters:
+    ///   - forRequestEventId: unique event identifier for which the completion handler is registered; should not be empty
+    ///   - completionHandler: the completion handler that needs to be registered, should not be nil
+    func registerCompletionHandler(forRequestEventId: String, completion: (([EdgeEventHandle]) -> Void)?) {
```
nporter-adbe, line 37: Is it possible to potentially register a completion handler twice for a single request id? I don't think there's a use case for this but I wonder if we should be safe and check if an existing completion handler is already mapped to the request id.

nporter-adbe (approved):

Sources/EdgeNetworkHandlers/CompletionHandlersManager.swift
emdobrin: Since this API is internal and used when sendEvent is called and a new Event is created, I think it is safe to say that there will not be multiple completion handlers associated with the same request id.

emdobrin:
@kevinlind let me now if you have any additional comments, I would like to merge by the end of the day.

PR #113, opened 2020-11-30
Update error network response handling

nporter-adbe:
This PR updates the error format based on the spec: https://git.corp.adobe.com/pages/experience-edge/konductor/#/apis/errors?id=errors-hierarchy

However, we also add the eventIndex property to the responses.

It also handles 207 with at least one error in the body.


Sources/EdgeNetworkHandlers/EdgeEventWarning.swift
```diff
@@ -0,0 +1,55 @@
...
+// OF ANY KIND, either express or implied. See the License for the specific language
+// governing permissions and limitations under the License.
+//
+
+import Foundation
+
+// MARK: - EdgeEventWarning
+
+/// Messages from upstreams which could signal that a partial processing of the request occurred
+struct EdgeEventWarning: Codable {
```
emdobrin, line 18: Based on https://github.com/adobe/aepsdk-edge-ios/pull/110 we would like to expose EdgeEventError to the caller in the completion handler and I think it might become too complicated for the developer to distinguish between the two. Did you also consider keeping them under the same class as until now, so EdgeEventError would have an optional report payload if provided by the server? This way the EdgeEventError is still in line with the AEP Error Response event.

Tests/FunctionalTests/NetworkResponseHandlerFunctionalTests.swift
```diff
@@ -43,8 +43,8 @@ class NetworkResponseHandlerFunctionalTests: FunctionalTestBase {
     func testProcessResponseOnError_WhenGenericJsonError_dispatchesEvent() {
         setExpectationEvent(type: FunctionalTestConst.EventType.EDGE, source: FunctionalTestConst.EventSource.ERROR_RESPONSE_CONTENT, expectedCount: 1)
         let jsonError = "{\n" +
-            "\"namespace\": \"global\",\n" +
-            "\"message\": \"Request to Data platform failed with an unknown exception\"" +
+            "\"type\": \"global\",\n" +
```
emdobrin, line 46: Let's use real examples like `"type":"https://ns.adobe.com/aep/errors/EXEG-0201-503"` instead of global. Applies for all occurrences of global.

Tests/FunctionalTests/NetworkResponseHandlerFunctionalTests.swift
```diff
@@ -576,9 +576,9 @@ class NetworkResponseHandlerFunctionalTests: FunctionalTestBase {
             "            ]}],\n" +
             "      \"errors\": [" +
             "        {\n" +
-            "          \"code\": \"personalization:2003\",\n" +
-            "          \"namespace\": \"personalization\",\n" +
-            "          \"message\": \"Failed to process personalization event\"\n" +
+            "          \"status\": 2003,\n" +
```
emdobrin, line 579: Status now represents the http status code, so this value is not possible. Let's update these tests with some accurate examples.

Tests/UnitTests/EdgeEventWarningTests.swift
```diff
@@ -0,0 +1,160 @@
...
+                            },
+                            {
+                              "eventIndex": 2,
+                              "type": "https://ns.adobe.com/aep/errors/EXEG-0204-202",
+                              "status": 202,
+                              "title": "test title 2",
+                              "report": {
+                                "cause": {
+                                  "message": "Cannot read related customer for device id: ...",
+                                  "code": 202
```
emdobrin, line 76: Can you also include some tests with unexpected parameters? We should capture the case where the server gets updated with new params, but the client runs current version.

Tests/UnitTests/EdgeNetworkServiceTests.swift
```diff
@@ -133,6 +133,29 @@ class EdgeNetworkServiceTests: XCTestCase {
 
         wait(for: [expectation], timeout: 0.5)
     }
+    
+    func testDoRequest_whenConnection_ResponseCode207_CallsCompletionTrue_AndResponseCallback_AndErrorCallback() {
```
emdobrin, line 137: Can you also add one e2e test with multi status, checking resulting response & error events?

emdobrin (changes requested):

Sources/EdgeNetworkHandlers/EdgeEventWarning.swift
nporter-adbe: I actually believe it would be more misleading/confusing to represent a warning from Konductor with a type that contains the name "Error," indicating it's an error rather than a warning. I do understand the perspective of having more types for the customer to understand could be confusing. However, I find it quite a bit more confusing to have one type that claims it's an error (by name) to actually represent both errors and warnings. We need to indicate what an error is and what is not an error. By separating these, we can clearly label one an error and one a warning. 

If we overload this type to represent both warnings and errors, I think we'll end up with more confusion, and it'll be easy to mistake an error for a warning.

emdobrin: So are you suggesting to also have a new event type that an extension should listen to?

nporter-adbe: Yeah, I think that may be a good idea. I believe for total clarity it will be best to be able to easily distinguish between warnings and events.

We can discuss this possibility further and I am happy to submit another PR to add the new event type and change the warning events from errors to warning type.

Tests/FunctionalTests/NetworkResponseHandlerFunctionalTests.swift
nporter-adbe: typo whoops.

Tests/UnitTests/EdgeEventWarningTests.swift
nporter-adbe: If I'm understanding this correctly, this isn't possible in Swift. We can only decode the properties/types we explicitly define in `EdgeEventWarning`. Anything that doesn't fit the definition in `EdgeEventWarning` will be discarded. If the format of Konductor warnings updates to append new properties we'll need to update the SDK to decode these.

emdobrin: That is correct, I think it is fine to discard extraneous keys until we introduce support for them, I am just suggesting to have a test to make sure we don't crash / drop the response if the server response gets updated.

nporter-adbe:
@emdobrin One question I have about how we want to handle 207's.

Currently, it will result in a duplicate error event being dispatched sine we invoke handleContent and handleError, both of which will dispatch error events. Do we think this is acceptable or do you have any suggestions?

```swift
else if responseCode == HttpResponseCodes.multiStatus.rawValue {
                    Log.debug(label: self.LOG_TAG,
                              "doRequest - Connection to Experience Edge was successful but encountered non-fatal errors. \(responseCode)")
                    self.handleContent(connection: connection,
                                       streaming: requestBody.meta?.konductorConfig?.streaming,
                                       responseCallback: responseCallback)
                    self.handleError(connection: connection, responseCallback: responseCallback)
                    responseCallback.onComplete()
                    completion(true) // non-fatal error, don't retry
}
```

nporter-adbe:
@emdobrin Based on our meeting today, it looks like no additional changes should need to be added for this PR as we'll be publishing warnings and errors via the same event type and source. Let me know if you agree.

However, I still have the question above. https://github.com/adobe/aepsdk-edge-ios/pull/113#issuecomment-745581238

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -178,9 +178,8 @@ class NetworkResponseHandler {
     /// - Parameters:
```
emdobrin, line 178: Let's rephrase to ->  `It also logs each error json with the log level error.`

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
```diff
@@ -737,6 +738,67 @@ class AEPEdgeFunctionalTests: FunctionalTestBase {
...
+                                                                response: HTTPURLResponse(url: exEdgeInteractUrl,
+                                                                                          statusCode: 207,
+                                                                                          httpVersion: nil,
+                                                                                          headerFields: nil),
+                                                                error: nil)
+        setNetworkResponseFor(url: exEdgeInteractUrlString, httpMethod: HttpMethod.post, responseHttpConnection: responseConnection)
+
+        setExpectationNetworkRequest(url: exEdgeInteractUrlString, httpMethod: HttpMethod.post, expectedCount: 1)
+        setExpectationEvent(type: FunctionalTestConst.EventType.EDGE, source: FunctionalTestConst.EventSource.REQUEST_CONTENT, expectedCount: 1) // the send event
+        setExpectationEvent(type: FunctionalTestConst.EventType.EDGE, source: FunctionalTestConst.EventSource.ERROR_RESPONSE_CONTENT, expectedCount: 3) // 2 error events (one from the onSuccess callback, one from onError callback), then one warning event.
```
emdobrin, line 757: should expect 2

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
```diff
@@ -737,6 +738,67 @@ class AEPEdgeFunctionalTests: FunctionalTestBase {
 
         self.assertNetworkRequestsCount()
     }
+
+    func testSendEvent_multiStatusResponse_dispatchesEvents() {
```
emdobrin, line 742: Can we include one test with fatal error too with error details at the top level in the json.

emdobrin (commented):
Looks good, added my comments in-line

Sources/EdgeNetworkHandlers/EdgeEventError.swift
```diff
@@ -12,18 +12,35 @@
...
 
-    /// Error code info
-    let code: String?
+// MARK: - EdgeErrorReport
+struct EdgeErrorReport: Codable, Equatable {
+    // An array of errors represented as strings
+    let errors: [String]?
+
+    /// Request ID corresponding to the error
+    let requestId: String?
```
emdobrin, line 42: We already include the requestId we have on the client, not sure if it is required to decode it from here.

Sources/EdgeNetworkHandlers/EdgeNetworkService.swift
```diff
@@ -120,6 +119,14 @@ class EdgeNetworkService {
                 } else if self.recoverableNetworkErrorCodes.contains(responseCode) {
                     Log.debug(label: self.LOG_TAG, "doRequest - Connection to Experience Edge returned recoverable error code \(responseCode)")
                     completion(false) // failed, but recoverable so retry
+                } else if responseCode == HttpResponseCodes.multiStatus.rawValue {
+                    Log.debug(label: self.LOG_TAG,
+                              "doRequest - Connection to Experience Edge was successful but encountered non-fatal errors. \(responseCode)")
```
emdobrin, line 124: nit: ... non-fatal errors/warnings. Response code: \(responseCode).

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
```diff
@@ -737,6 +738,108 @@ class AEPEdgeFunctionalTests: FunctionalTestBase {
...
+                          "title" : "Unprocessable Entity",
+                          "detail": "Invalid request (report attached). Please check your input and try again.",
+                          "report": {
+                            "errors": [
+                              "Allowed Adobe version is 1.0 for standard 'Adobe' at index 0",
+                              "Allowed IAB version is 2.0 for standard 'IAB TCF' at index 1",
+                              "IAB consent string value must not be empty for standard 'IAB TCF' at index 1"
+                            ],
+                            "requestId": "0f8821e5-ed1a-4301-b445-5f336fb50ee8",
+                            "orgId": "53A16ACB5CC1D3760A495C99@AdobeOrg"
```
emdobrin, line 807: let's mock this: test@AdobeOrg

Tests/UnitTests/EdgeEventErrorTests.swift
```diff
@@ -0,0 +1,78 @@
...
+import XCTest
+
+class EdgeEventErrorTests: XCTestCase {
+
+    override func setUp() {
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+        continueAfterFailure = false // fail so nil checks stop execution
+    }
+
+    func testCanDecode_eventError_allParams() {
```
emdobrin, line 23: Let's add decode tests for fatal errors too.

emdobrin (commented):
Looks good to me with few small suggestions

Sources/EdgeNetworkHandlers/EdgeEventError.swift
nporter-adbe: I'd rather match the incoming schema as best as we can rather than omit properties, even if not used.

nporter-adbe:
@emdobrin Going to merge this end of today to begin working on the retry interval task. Let me know if you've got any additional concerns.

emdobrin (approved):

Issue #123, opened 2021-01-05
Respect retry interval in response header


nporter-adbe:
We need to base our retry interval based on the response header retry interval. This can be accomplished once https://github.com/adobe/aepsdk-core-ios/issues/485 is supported.

emdobrin:
Part of https://github.com/adobe/aepsdk-edge-ios/issues/125

PR #129, opened 2021-01-08
Respect retry-after header

nporter-adbe:
Addresses: https://github.com/adobe/aepsdk-edge-ios/issues/123

Also fixes an issue where CI was using an old version of Cocoapods that is not compatible with the latest version of AEPAssurance. The fix is to define a pod version in a Gemfile, then update our CI commands to run with `bundle exec`.


Gemfile
```diff
@@ -0,0 +1,3 @@
+source 'https://rubygems.org'
+
+gem 'cocoapods', '= 1.10.0'
```
nporter-adbe, line 3: This is required to fix the build issue seen here: https://app.circleci.com/pipelines/github/adobe/aepsdk-edge-ios/331/workflows/6f354a9c-01b8-4341-92fb-8b7be957f743/jobs/354

Sources/EdgeNetworkHandlers/EdgeNetworkService.swift
```diff
@@ -71,15 +71,19 @@ class EdgeNetworkService {
...
     ///   - requestHeaders: request HTTP headers
     ///   - responseCallback: `ResponseCallback` to be invoked once the server response is received
-    ///   - completion: a closure that is invoked with true if the hit should not be retried, false if the hit should be retried
-    func doRequest(url: URL, requestBody: EdgeRequest, requestHeaders: [String: String]? = [:], responseCallback: ResponseCallback, completion: @escaping (Bool) -> Void) {
+    ///   - completion: a closure that is invoked with true if the hit should not be retried, false if the hit should be retried, along with the time interval that should elapse before retrying the hit
+    func doRequest(url: URL,
+                   requestBody: EdgeRequest,
+                   requestHeaders: [String: String]? = [:],
+                   responseCallback: ResponseCallback,
+                   completion: @escaping (Bool, TimeInterval) -> Void) {
```
emdobrin, line 79: We can use `TimeInterval?`

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -33,6 +33,12 @@ class EdgeHitProcessor: HitProcessing {
         self.readyForEvent = readyForEvent
     }
 
+    // MARK: HitProcessing
+
+    func retryInterval(for entity: DataEntity) -> TimeInterval {
+        return retryIntervalMapping[entity.uniqueIdentifier] ?? EdgeConstants.NetworkKeys.RETRY_INTERVAL
```
emdobrin, line 39: Can you move this to EdgeConstants.Defaults.RETRY_INTERVAL ?

Sources/EdgeNetworkHandlers/EdgeNetworkService.swift
```diff
@@ -110,34 +114,39 @@ class EdgeNetworkService {
...
                     Log.debug(label: self.LOG_TAG, "doRequest - Collect connection to Experience Edge was successful.")
                     responseCallback.onComplete()
-                    completion(true) // successful request, return true
+                    completion(true, 0.0) // successful request, return true
                 } else if self.recoverableNetworkErrorCodes.contains(responseCode) {
                     Log.debug(label: self.LOG_TAG, "doRequest - Connection to Experience Edge returned recoverable error code \(responseCode)")
-                    completion(false) // failed, but recoverable so retry
+                    let retryHeader = connection.responseHttpHeader(forKey: EdgeConstants.NetworkKeys.HEADER_KEY_RETRY_AFTER)
+                    var retryInterval = EdgeConstants.NetworkKeys.RETRY_INTERVAL
+                    if let retryHeader = retryHeader, let retryAfterInterval = TimeInterval(retryHeader) {
```
emdobrin, line 127:
```
var retryAfterInterval = nil
if let retryHeader = connection.responseHttpHeader(forKey: EdgeConstants.NetworkKeys.HEADER_KEY_RETRY_AFTER) 
    retryAfterInterval = TimeInterval(retryHeader)
}
completion(false, retryInterval)

File: Tests/UnitTests/EdgeHitProcessorTests.swift [hunk 5]
```diff
@@ -241,18 +241,27 @@ class EdgeHitProcessorTests: XCTestCase {
...
             let error = EdgeEventError(title: "test-title", detail: nil, status: code, type: "test-type", eventIndex: 0, report: nil)
             let edgeResponse = EdgeResponse(requestId: "test-req-id", handle: nil, errors: [error], warnings: nil)
             let responseData = try? JSONEncoder().encode(edgeResponse)
 
-            mockNetworkService?.connectAsyncMockReturnConnection = HttpConnection(data: responseData, response: HTTPURLResponse(url: URL(string: "adobe.com")!, statusCode: code, httpVersion: nil, headerFields: nil), error: nil)
+            mockNetworkService?.connectAsyncMockReturnConnection = HttpConnection(data: responseData,
+                                                                                  response: HTTPURLResponse(url: URL(string: "adobe.com")!,
+                                                                                  statusCode: code,
+                                                                                  httpVersion: nil,
+                                                                                  headerFields: ["Retry-After": retryValueTuple.0]),
```
emdobrin, line 256: Is this header read by the network service as int from the server response and stored as string in the HttpConnection or do we need to add support for other header types?

Sources/EdgeNetworkHandlers/EdgeNetworkService.swift
emdobrin, line 127: It seems like the Retry-After header can be: 
```
<http-date>
A date after which to retry. See the Date header for more details on the HTTP date format.
<delay-seconds>
A non-negative decimal integer indicating the seconds to delay after the response is received.
```
As discussed with the service team, this will be in int format, but let's clarify that in a comment here and a test case that we don't fail when date is returned.

Ref https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

Tests/UnitTests/EdgeNetworkServiceTests.swift
```diff
@@ -338,12 +419,13 @@ class EdgeNetworkServiceTests: XCTestCase {
 
         // test
         let mockHttpConnection = HttpConnection(data: stringResponseBody.data(using: .utf8),
-                                                response: HTTPURLResponse(url: url, statusCode: 503, httpVersion: nil, headerFields: nil),
+                                                response: HTTPURLResponse(url: url, statusCode: 503, httpVersion: nil, headerFields: ["Retry-After": "60"]),
```
emdobrin, line 422: We should only set retryInterval when no call to onComplete and should retry.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -120,8 +127,11 @@ class EdgeHitProcessor: HitProcessing {
         networkService.doRequest(url: url,
                                  requestBody: edgeHit.request,
                                  requestHeaders: headers,
-                                 responseCallback: callback,
-                                 completion: completion)
+                                 responseCallback: callback) { [weak self] success, retryInterval in
+            // remove any retry interval if success, add retry interval if failed
+            self?.retryIntervalMapping[entityId] = success ? nil : retryInterval
```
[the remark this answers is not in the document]
nporter-adbe: That doesn't cover the case where we need to remove a hit that was once retried (has an entry in retry interval map) and needs to be cleaned up.

Sources/EdgeNetworkHandlers/EdgeNetworkService.swift
nporter-adbe: Yes, it's just a double under the hood.

Makefile
```diff
@@ -9,23 +9,23 @@ SIMULATOR_ARCHIVE_PATH = ./build/ios_simulator.xcarchive/Products/Library/Framew
 IOS_ARCHIVE_PATH = ./build/ios.xcarchive/Products/Library/Frameworks/
 
 setup:
-	(pod install)
-	(cd SampleApps/$(APP_NAME) && pod install)
+	(bundle exec pod install)
```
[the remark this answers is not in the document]
nporter-adbe: Yeah, we'll probably need to separate the commands out to ci vs local commands. We can do that in a separate PR. I'm just doing the minimal possible to get CI to pass given the changes to Assurance.

Sources/EdgeNetworkHandlers/EdgeNetworkService.swift
nporter-adbe: I don't see any need to add a test case for a scenario that will not be returned by the server and one that we explicitly do not support. I've added a clarifying comment to the parsing of the header.

Tests/UnitTests/EdgeHitProcessorTests.swift
nporter-adbe: HTTP headers are string/string key values, I don't believe there is any need to support other values. If you look at other networking libraries you can see that only [String: String] headers are supported.

https://stackoverflow.com/questions/34152142/possible-types-of-a-http-header-value#:~:text=The%20type%20of%20a%20value,string%20is%20actually%20a%20string.

Sources/EdgeNetworkHandlers/EdgeNetworkService.swift
nporter-adbe: I'm exactly sure what the suggestion is. But the current code snippet above doesn't compile.  Mainly due to the fact that `TimeInterval` is a typealias of a `Double`, so it can't take the value of nil and the completion `(Bool, TimerInterval)` can't support passing in `(boolValue, nil)`, so we must provide a default value.

Tests/UnitTests/EdgeNetworkServiceTests.swift
nporter-adbe: Looks like the test name may be wrong, but the error code of 503 is recoverable therefore we should retry this hit, and subsequently set the retry interval. Is this correct?

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -120,8 +127,11 @@ class EdgeHitProcessor: HitProcessing {
         networkService.doRequest(url: url,
                                  requestBody: edgeHit.request,
                                  requestHeaders: headers,
-                                 responseCallback: callback,
-                                 completion: completion)
+                                 responseCallback: callback) { [weak self] success, retryInterval in
+            // remove any retry interval if success or has default interval value, add retry interval if failed
+            self?.entityRetryIntervalMapping[entityId] = success || retryInterval == EdgeConstants.Defaults.RETRY_INTERVAL ? nil : retryInterval
```
emdobrin, line 132: We should explicitly call the removeValue API to remove the pre-existing mapping in case the internal implementation of ThreadSafeDictionary changes in the future. Considering TimeInterval is optional, you can have something like this:
```
if success {
     // remove retry interval if success
     _ = self?.entityRetryIntervalMapping.removeValue(forKey: entityId)
} else if retryInterval != nil {
     // add retry interval if failed
     self?.entityRetryIntervalMapping[entityId] = retryInterval
}
...
```

Tests/UnitTests/EdgeHitProcessorTests.swift
nporter-adbe: It's strange to me they define it as `[AnyHashable: Any]` because they only allow you to read and index them as strings.
https://developer.apple.com/documentation/foundation/httpurlresponse/3240613-value

Sources/EdgeNetworkHandlers/EdgeNetworkService.swift
nporter-adbe: As stated above `TimeInterval` cannot be `nil`. It is a type alias of `Double` which cannot take the value of `nil`.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
nporter-adbe: The existing line of code already does this. Setting a value in a dictionary to nil is the equivalent to removing it explicitly. We would never modify the implementation of ThreadSafeDIctionary to make it not behave like a standard Swift dict when used through the subscript.

Tests/UnitTests/EdgeNetworkServiceTests.swift
nporter-adbe: Went ahead and updated this test name.

Sources/EdgeNetworkHandlers/EdgeNetworkService.swift
nporter-adbe: Sorry, I misunderstood this context. I read it as "Can we use a `TimeInterval`?", Not can we use an optional time interval.

As stated in another comment, it's not possible since TimeInterval is a Double which cannot take on nil.

Makefile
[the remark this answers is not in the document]
nporter-adbe: Also, the bundle exec commands still work locally (for me at least). It could be a good switch to bundle exec to ensure we're all running the sam eversion of cocoapods.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
emdobrin: Since it is out of this extension's control I would not rely that there will be no changes. The APIs can remain the same and the underlying data structure can be different. Otherwise why do we have a removeValue API?

Makefile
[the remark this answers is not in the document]
emdobrin: It requires you to handle the bundler installed locally, which shouldn't be required in my opinion. Can we add bundle exec in the circleci config or update the makefile for ci? 
@kevinlind what do you think?

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
nporter-adbe: It actually is somewhat in control of the extension based on the definition of `ThreadSafeDictionary<String, TimeInterval>()` 

The definition explicitly states it cannot hold `nil` values, thus setting any key to `nil` can have no other side-effect than to remove it from the dictionary if present.

Makefile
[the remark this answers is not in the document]
nporter-adbe: Updated the CI script with CI only commands.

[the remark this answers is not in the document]
kevinlind: I see Nick already updated the Makefile, so that's fine. Don't need to change it now. Though, I  would be okay with investigating using bundle exec in the Makefile. The setup target could verify the bundler is installed and install it if needed. I do find it annoying when we get out of sync with our CocoaPods versions and `make setup` causes a diff.

emdobrin (approved):

PR #131, opened 2021-01-13
Add additional test and remove dead code

nporter-adbe:
This adds a real fatal error response test I received from Konductor by providing an invalid config id. This PR doesn't fix any bugs, but rather adds a new test and does some minor clean up.

I also took a minute to take another look at some of the decoding code, and realized that the decoding within the `guard` is redundant, and checking if `errors` is nil is sufficient. In the case a generic fetal error is returned, it will end up decoding into a `nil` `EdgeErrorResponse`, then when we check `redgeErrorResponse.errors` we'll realize that we need to decode it as a generic error.

### DIFF
--- a/Sources/EdgeNetworkHandlers/EdgeResponse.swift
+++ b/Sources/EdgeNetworkHandlers/EdgeResponse.swift
@@ -17,7 +17,7 @@ import Foundation
 struct EdgeResponse: Codable {
 
     /// The request identifier associated with this response
-    let requestId: String?
+    let requestId: String
 
     /// List of event handles received from the Experience Edge  Network
     let handle: [EdgeEventHandle]?
--- a/Sources/EdgeNetworkHandlers/EdgeResponse.swift
+++ b/Sources/EdgeNetworkHandlers/EdgeResponse.swift
@@ -17,7 +17,7 @@ import Foundation
 struct EdgeResponse: Codable {
 
     /// The request identifier associated with this response
-    let requestId: String
+    let requestId: String?
 
     /// List of event handles received from the Experience Edge  Network
     let handle: [EdgeEventHandle]?
--- a/Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
+++ b/Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
@@ -90,7 +90,8 @@ class NetworkResponseHandler {
         guard let data = jsonError.data(using: .utf8) else { return }
         Log.debug(label: LOG_TAG, "processResponseOnError - Processing server error response:\n \(jsonError), request id \(requestId)")
 
-        if let edgeResponse = try? JSONDecoder().decode(EdgeResponse.self, from: data), edgeResponse.errors != nil {
+        if let edgeResponse = try? JSONDecoder().decode(EdgeResponse.self, from: data) {
+            guard edgeResponse.errors != nil else { return }
             // this is an error coming from Konductor, read the error from the errors node
             dispatchEventErrors(errorsArray: edgeResponse.errors, requestId: requestId)
         } else if let edgeErrorResponse = try? JSONDecoder().decode(EdgeEventError.self, from: data) {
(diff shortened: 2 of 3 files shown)


Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -91,16 +91,8 @@ class NetworkResponseHandler {
...
-                Log.debug(label: LOG_TAG, "processResponseOnError - Processing server error response:\n \(jsonError), request id \(requestId)")
-                dispatchEventErrors(errorsArray: [edgeErrorResponse], requestId: requestId)
-            } else {
-                Log.warning(label: LOG_TAG,
-                            "processResponseOnError - The conversion to JSON failed for server error response: \(jsonError), request id \(requestId)")
-            }
-
+            Log.warning(label: LOG_TAG,
+                        "processResponseOnError - The conversion to JSON failed for server error response: \(jsonError), request id \(requestId)")
             return
```
emdobrin, line 96: It looks like we account for generic errors in lines L107->L115, so we should first attempt to decode to EdgeResponse, then EdgeEventError, without returning here.

nporter-adbe:
Elected to just re-write the entire `processResponseOnError` function.

Sources/EdgeNetworkHandlers/EdgeResponse.swift
```diff
@@ -17,7 +17,7 @@ import Foundation
 struct EdgeResponse: Codable {
 
     /// The request identifier associated with this response
-    let requestId: String?
+    let requestId: String
```
emdobrin, line 20: Why was this change required?

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -88,39 +88,20 @@ class NetworkResponseHandler {
...
-
-            return
-        }
-
         Log.debug(label: LOG_TAG, "processResponseOnError - Processing server error response:\n \(jsonError), request id \(requestId)")
 
-        // Note: if the Edge error doesn't have an eventIndex it means that this error is a generic request error,
-        // otherwise it is an event specific error. There can be multiple errors returned for the same event
-        if edgeErrorResponse.errors != nil {
+        if let edgeResponse = try? JSONDecoder().decode(EdgeResponse.self, from: data), edgeResponse.errors != nil {
```
emdobrin, line 93: I think we should move edgeResponse.errors != nil inside the if or just let it fast fail in dispatchEventErrors. If we were able to decode to EdgeResponse we should not try to decode to EdgeEventError.
```
if let edgeResponse = try? JSONDecoder().decode(EdgeResponse.self, from: data) { ...
else if let edgeErrorResponse = try? JSONDecoder().decode(EdgeEventError.self, from: data) { ...
```

Sources/EdgeNetworkHandlers/EdgeResponse.swift
nporter-adbe: Marking the `requestId` as non-nil (required) was needed because if we would attempt to decode `Data` that represented an `EdgeEventError` to a `EdgeResponse` would end up with a non-nil `EdgeResponse` but with all nil properties.

Here is an example:
```swift
let edgeError = EdgeEventError(...)
let edgeErrorData = try! JSONEncoder().encode(edgeError)
let edgeResponse = try! JSONDecoder().decode(EdgeResponse.self, edgeErrorData)
XCTAssertNil(edgeResponse) 
```
Originally this assertion would have failed, it would have just initialized an empty `EdgeResponse` with all nil properties (since they were all optional). Now that we require there to be a `requestId` at the top level the decoding will fail, and return nil as we expect.

nporter-adbe: This change essentially allow us to better match a response we'll get back from Kondcutor as `requestId` is a guaranteed field, thus allowing us to write cleaner code when it comes to decoding response vs error data.

emdobrin: Can we fail the decoding when all properties are nil/not set?

nporter-adbe: No, I looked into that. You could probably create a `isEmpty` flag that looks to see if all the properties are nil but I don't believe that's good practice in Swift. We should be marking all our decoded/encoded properties as nil or non-nil based on the API spec. However, some of the spec is a bit foggy to me as to whats optional/required.

AFAIK, there is no situation where we will receive a valid response (non-fatal) that doesn't contain a non-nil `requestId`. I'm having a hard time double-checking that as Konductors docs are giving back a 503.

PR #132, opened 2021-01-13
Enabled skipped Identity test 

emdobrin:


Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -22,6 +22,7 @@ class EdgeHitProcessor: HitProcessing {
     private var getSharedState: (String, Event?) -> SharedStateResult?
     private var readyForEvent: (Event) -> Bool
     private var entityRetryIntervalMapping = ThreadSafeDictionary<String, TimeInterval>()
+    var batchLimit: Int = 0
```
nporter-adbe, line 25: let's make this a `let`

nporter-adbe (approved):
one minor comment

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
nporter-adbe: oh actually, the protocol requires a setter. ignore this comment.

emdobrin: I am getting this error `Type 'EdgeHitProcessor' does not conform to protocol 'HitProcessing' Candidate is not settable, but protocol requires it`. I see `var batchLimit: Int { get set} ` in the protocol. I guess we only need the getter, do we use the setter somewhere else?

nporter-adbe: Yeah my bad, I forgot we require a setting on the batch limit. I believe this is mainly to support if the batch limit is updated in Analytics via config updates.

PR #138, opened 2021-01-15
Make public APIs ObjC Compatible

nporter-adbe:
Our current APIs do not allow for creating an Event or sending an Event in ObjC.

### DIFF
--- a/Sources/ExperienceEvent.swift
+++ b/Sources/ExperienceEvent.swift
@@ -16,23 +16,23 @@ import Foundation
 public class ExperienceEvent: NSObject {
 
     /// XDM formatted data, use an `XDMSchema` implementation for a better XDM data injestion and format control
-    public let xdm: [String: Any]?
+    @objc public let xdm: [String: Any]?
 
     /// Optional free-form data associated with this event
-    public let data: [String: Any]?
+    @objc public let data: [String: Any]?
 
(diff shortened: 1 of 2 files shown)


Sources/ExperienceEvent.swift
```diff
@@ -16,23 +16,23 @@ import Foundation
...
-    public let datasetIdentifier: String?
+    @objc public let datasetIdentifier: String?
 
     /// Initialize an Experience Event with the provided event data
     /// - Parameters:
     ///   - xdm:  XDM formatted data for this event, passed as a raw XDM Schema data dictionary.
     ///   - data: Any free form data in a [String : Any] dictionary structure.
     ///   - datasetIdentifier: The Experience Platform dataset identifier where this event should be sent to; if not provided, the default dataset identifier set in the Edge configuration is used
-    public init(xdm: [String: Any], data: [String: Any]? = nil, datasetIdentifier: String? = nil) {
+    @objc public init(xdm: [String: Any], data: [String: Any]? = nil, datasetIdentifier: String? = nil) {
```
emdobrin, line 35: Any changes we need to do to XDMSchema or XDMFormatters? 
EdgeEventHandle seems correct, but please test with that too.

emdobrin (approved):

kevinlind (approved):

PR #140, opened 2021-01-19
Release automation

nporter-adbe:
This PR ports over release automation from the core repo. It allows us to automatically push our pod, validate pod usage, and validate SPM usage. We run these scripts via a button click.

https://github.com/adobe/aepsdk-core-ios/actions?query=workflow%3ARelease

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -8,11 +8,6 @@ on:
         required: true
         default: '1.0.0'
 
-      action_verification:
-        description: 'Pod/SPM verification ("no" to skip)'
-        required: true
-        default: 'yes'
-
       action_tag:
         description: 'create tag ("no" to skip)'
         required: true
@@ -44,14 +39,14 @@ jobs:
         make check-version VERSION=${{ github.event.inputs.tag }}
 
     - name: SPM integration test
-      if: ${{ github.event.inputs.action_verification == 'yes' }}
+      if: ${{ github.event.inputs.action_tag == 'yes' }}
       run: |
         set -eo pipefail
         echo SPM integration test starts:
         make test-SPM-integration
 
     - name: podspec file verification
-      if: ${{ github.event.inputs.action_verification == 'yes' }}
+      if: ${{ github.event.inputs.action_tag == 'yes' }}
       run: |
         set -eo pipefail
         echo podspec file verification starts:
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,77 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'tag/version'
+        required: true
+        default: '1.0.0'
+
+      action_verification:
+        description: 'Pod/SPM verification ("no" to skip)'
+        required: true
+        default: 'yes'
+
+      action_tag:
+        description: 'create tag ("no" to skip)'
+        required: true
+        default: 'yes'
+
+      release_AEPEdge:
+        description: 'release AEPEdge to Cocoapods ("no" to skip)'
+        required: true
+        default: 'yes'
+
+jobs:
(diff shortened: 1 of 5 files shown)


.github/release.yml
```diff
@@ -0,0 +1,77 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'tag/version'
+        required: true
+        default: '1.0.0'
+
+      action_verification:
```
yangyansong-adbe, line 11: https://github.com/adobe/aepsdk-core-ios/pull/505/files
Can we also port above changes to skip the 'verification' step only if 'create tag' is skipped?

Script/version.sh
```diff
@@ -0,0 +1,30 @@
...
+
+echo "Target version - ${BLUE}$1${NC}"
+echo "-----------------AEPEdge------------------"
+PODSPEC_VERSION_IN_AEPEdge=$(pod ipc spec AEPEdge.podspec | jq '.version' | tr -d '"')
+echo "Local podspec version - ${BLUE}${PODSPEC_VERSION_IN_AEPEdge}${NC}"
+# SOUCE_CODE_VERSION_IN_AEPEdge=$(cat ./AEPSignal/Sources/SignalConstants.swift | egrep '\s*EXTENSION_VERSION\s*=\s*\"(.*)\"' | ruby -e "puts gets.scan(/\"(.*)\"/)[0] " | tr -d '"')
+# echo "Souce code version - ${BLUE}${SOUCE_CODE_VERSION_IN_AEPEdge}${NC}"
+
+# if [[ "$1" == "$PODSPEC_VERSION_IN_AEPEdge" ]] && [[ "$1" == "$SOUCE_CODE_VERSION_IN_AEPEdge" ]]; then
+if [[ "$1" == "$PODSPEC_VERSION_IN_AEPEdge" ]]; then
```
yangyansong-adbe, line 24: https://github.com/adobe/aepsdk-core-ios/blob/c9abdd670e602c515889a94b189dab6b18887291/Script/version.sh#L43 
Should we also check the version in source code?

yangyansong-adbe (approved):

Makefile
```diff
@@ -66,3 +66,12 @@ lint-autocorrect:
 
 lint:
 	(swiftlint lint Sources SampleApps/AEPCommerceDemoApp)
+
+check-version:
+	(sh ./script/version.sh $(VERSION))
```
emdobrin, line 71: should be uppercase Script `./Script/version.sh`
same for the other checks below

Script/test-SPM.sh
```diff
@@ -0,0 +1,70 @@
...
+        .iOS(.v10)
+    ],
+    products: [
+        .library(
+            name: \"TestProject\",
+            targets: [\"TestProject\"]
+        )
+    ],
+    dependencies: [
+        .package(url: "https://github.com/adobe/aepsdk-core-ios.git", from: "3.0.0-beta.3"),
```
emdobrin, line 32: we should make core version configurable as well

nporter-adbe: Once core is released I plan to change this to `from: "3.0.0"` that way it always pulls the latest version of core. I'm not sure we should allow us to configure anything other than the latest version of core when running this script. otherwise, if the wrong version of core is used when running the release automation we could get invalid results based on the latest release of core.

yangyansong-adbe: @nporter-adbe 
Do we really need to add the AEPCore dependency here? https://github.com/adobe/aepsdk-edge-ios/blob/f96a7bf61a3b4bdbf234989139d8cf9a64004606/Package.swift#L31  (AEPEdge -> AEPCore -> AEPRulesEngine, AEPService), I think SPM will find the right AEPCore Package path from above code.

PR #142, opened 2021-01-19
Clear client-side store on privacy opt-out 

emdobrin:


Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -23,6 +23,7 @@ class NetworkResponseHandler {
 
     // the order of the request events matter for matching them with the response events
     private var sentEventsWaitingResponse = ThreadSafeDictionary<String, [String]>()
+    static var currentPrivacyStatus: PrivacyStatus = EdgeConstants.DEFAULT_PRIVACY_STATUS
```
nporter-adbe, line 26: Is it possible that `currentPrivacyStatus` could be written and read at the same time? Given that it is set from Edge.extension and read from within NetworkResponseHandler. handleStoreEventHandle.

Sources/Edge.swift
```diff
@@ -31,6 +30,7 @@ public class Edge: NSObject, Extension {
 
     public required init(runtime: ExtensionRuntime) {
         self.runtime = runtime
+        NetworkResponseHandler.currentPrivacyStatus = EdgeConstants.DEFAULT_PRIVACY_STATUS
```
nporter-adbe, line 33: This is a bit redundant with the definition in the NetworkResponseHandler.
`static var currentPrivacyStatus: PrivacyStatus = EdgeConstants.DEFAULT_PRIVACY_STATUS`

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
nporter-adbe: One suggestion would be to define a `getPrivacyStatus` function in Edge, then inject it into the NetworkResponseHandler.

```swift
// Edge.swift
init() {
   // inject closure
   networkResponseHandler = NetworkResponseHandler(getPrivacyStatus: getPrivacyStatus)
}

func getPrivacyStatus() -> PrivacyStatus {
   // read and return privacy status from shared state
}
```

```swift
// NetworkResponseHandler.swift
init(getPrivacyStatus: () -> PrivacyStatus) {
   // store the closure, then use where needed
}
```

Then you can use `getPrivacyStatus()` in `NetworkResponseHandler. handleStoreEventHandle` to read the privacy status without worry about threading concerns on the static variable. This would also increase testability as we no longer rely on a static variable being used within an instance.

The above might not compile 100%, but that's the idea I have.

emdobrin: Updated, see below.

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -263,6 +263,13 @@ class NetworkResponseHandler {
     /// If handle is of type "state:store" persist it to Data Store
     /// - Parameter handle: current `EventHandle` to store
     private func handleStoreEventHandle(handle: EdgeEventHandle) {
+        let storeResponsePayloadManager = StoreResponsePayloadManager(EdgeConstants.DataStoreKeys.STORE_NAME)
+        // handling async scenario when in-flight response is handled after the privacy status update was handled by the Edge extension
+        if Edge.currentPrivacyStatus == .optedOut {
+            storeResponsePayloadManager.deleteAllStorePayloads()
```
pfransenadb, line 274: Do we need to delete here or just return?  Seems like the delete is overkill, because the actual delete would've already been processed by handleConfigurationResponse.

pfransenadb (commented):
Minor question about necessity of re-clear on async case.

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
emdobrin: not necessary indeed

nporter-adbe (approved):

pfransenadb (approved):

PR #144, opened 2021-01-19
staging -> main (v1.0.0)

nporter-adbe:


nporter-adbe:
Before merging this PR, we'll probably want to update the Core dependencies in our Podfile.lock to the GA versions.

Podfile
```diff
@@ -8,42 +8,42 @@ workspace 'AEPEdge'
...
 end
 
 target 'AEPDemoAppSwiftUI' do
-  pod 'AEPCore', :git => 'git@github.com:adobe/aepsdk-core-ios.git', :branch => 'main'
-  pod 'AEPServices', :git => 'git@github.com:adobe/aepsdk-core-ios.git', :branch => 'main'
-  pod 'AEPRulesEngine', :git => 'git@github.com:adobe/aepsdk-rulesengine-ios.git', :branch => 'main'
-  pod 'AEPLifecycle', :git => 'git@github.com:adobe/aepsdk-core-ios.git', :branch => 'main'
-  pod 'AEPIdentity', :git => 'git@github.com:adobe/aepsdk-core-ios.git', :branch => 'main'
-  pod 'AEPSignal', :git => 'git@github.com:adobe/aepsdk-core-ios.git', :branch => 'main'
+  pod 'AEPCore', :git => 'git@github.com:adobe/aepsdk-core-ios.git', :branch => 'dev-v3.0.0'
```
emdobrin, line 30: We should update these dependencies once Core extensions landed in main

AEPEdge.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "AEPEdge"
-  s.version          = "1.0.0-beta.2"
+  s.version          = "1.0.0"
```
emdobrin, line 3: @shalehaha @nporter-adbe I think this podspec should declare a dependency on ` s.dependency 'AEPCore', '>= 3.0' ` 
also, let's update swift version to 5.1 here as well

nporter-adbe: @emdobrin I updated the podspec to 5.1 for the swift version, but I'm not seeing the point to enforce >= 3.0 as there will never be a version less than 3.0 for AEPCore.

emdobrin: I think it is safe we can remove AEPServices since it is part of the AEPCore spec

nporter-adbe: updated.

PR #148, opened 2021-01-19
Update dependencies to GA

nporter-adbe:
Updates dependencies to GA versions.


Podfile
```diff
@@ -8,42 +8,42 @@ workspace 'AEPEdge'
 project 'AEPEdge.xcodeproj'
 
 target 'AEPEdge' do
-  pod 'AEPCore', :git => 'git@github.com:adobe/aepsdk-core-ios.git', :branch => 'dev-v3.0.0'
-  pod 'AEPServices', :git => 'git@github.com:adobe/aepsdk-core-ios.git', :branch => 'dev-v3.0.0'
-  pod 'AEPRulesEngine', :git => 'git@github.com:adobe/aepsdk-rulesengine-ios.git', :tag => 'main'
+  pod 'AEPCore'
+  pod 'AEPServices'
+  pod 'AEPRulesEngine'
```
emdobrin, line 13: Can we remove AEPServices and AEPRulesEngine and rely on AEPCore to bring those (same below)

nporter-adbe: Yeah, it's not required to specify them. My thought on including these directly is that it gives us a little more flexibility during development to pull in specific versions of services and rules engine if we want a specific version/branch for testing/dev purposes.

emdobrin (approved):

PR #160, opened 2021-01-20
Update compatibility layer in sample app

nporter-adbe:


emdobrin (approved):

emdobrin:
@nporter-adbe we might need to fix this into dev since the release tag was already created in main

PR #165, opened 2021-02-17
[AMSDK-11077] Read XDM shared state from IdentityEdge

nporter-adbe:
- Updates project to use IdentityEdge
- Reads XDM shared state from IdentityEdge instead of standard shade state from Identity
- Parses the shared state as an `IdentityMap`
- Updates some tests


Sources/EdgeNetworkHandlers/EdgeEventEncodable.swift
```diff
@@ -1,24 +0,0 @@
...
-// Unless required by applicable law or agreed to in writing, software distributed under
-// the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
-// OF ANY KIND, either express or implied. See the License for the specific language
-// governing permissions and limitations under the License.
-//
-
-import Foundation
-
-/// Encodable extension used by `EdgeEventHandle` and `EdgeEventError`
-extension Encodable {
```
nporter-adbe, line 16: This is a duplicate of what is in AEPServices. Sometimes it causes the compiler to get confused.

Sources/EdgeNetworkHandlers/IdentityMap.swift
```diff
@@ -54,6 +56,23 @@ struct IdentityMap {
     func getItemsFor(namespace: String) -> [IdentityItem]? {
         return items[namespace]
     }
+
+    /// Decodes a [String: Any] dictionary into an `IdentityMap`
+    /// - Parameter eventData: the event data representing `IdentityMap`
+    /// - Returns: an `IdentityMap` that is represented in the event data, nil if data is not in the correct format
+    static func from(eventData: [String: Any]) -> IdentityMap? {
```
nporter-adbe, line 63: Copied from IdentityEdge.

emdobrin: What if we pass the IdentityMap from XDM shared state to Konductor directly as dictionary instead of decoding it as IdentityMap? Edge doesn't need to know the IdentityMap structure anymore as the XDM Shared state should guarantee the correct XDM format based on the API contract.

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
```diff
@@ -34,16 +34,13 @@ class AEPEdgeFunctionalTests: FunctionalTestBase {
         continueAfterFailure = false
         FileManager.default.clearCache()
 
-        // hub shared state update for 2 extension versions (InstrumentedExtension (registered in FunctionalTestBase), Identity, Edge), Identity and Config shared state updates
-        setExpectationEvent(type: FunctionalTestConst.EventType.HUB, source: FunctionalTestConst.EventSource.SHARED_STATE, expectedCount: 4)
+        // hub shared state update for 2 extension versions (InstrumentedExtension (registered in FunctionalTestBase), IdentityEdge, Edge), IdentityEdge standard, IdentityEdge XDM and Config shared state updates
```
emdobrin, line 37: cc @lind I think we will not need the standard shared state for IdentityEdge anymore

Sources/EdgeNetworkHandlers/RequestBuilder.swift
```diff
@@ -23,8 +23,8 @@ class RequestBuilder {
     /// Control character used at the end of each response fragment. Response streaming is enabled when both `recordSeparator` and `lineFeed` are non nil.
     private var lineFeed: String?
 
-    /// The Experience Cloud ID to be sent with this request
-    var experienceCloudId: String?
+    /// The`IdentityMap` to be attached to the request
+    var identityMap: IdentityMap?
```
emdobrin, line 27: As in my other comment, I would prefer to keep the Identity map as raw dictionary

emdobrin:
@nporter-adbe we will also need to declare the IdentityEdge dependency in the podspec as well

Sources/EdgeNetworkHandlers/IdentityMap.swift
nporter-adbe: Ok, we can just delete IdentityMap from this repo.

nporter-adbe:
@emdobrin Adding identity edge to the podspec dependencies causes CI failure, we may need to add it later: https://app.circleci.com/pipelines/github/adobe/aepsdk-edge-ios/440/workflows/6609f1f2-7e57-4458-a2ac-765e36ca636d/jobs/469/parallel-runs/0/steps/0-106

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
kevinlind: Yes, I'm working on some changes which removes the regular shared state from IdentityEdge. I'll have a PR later this week.

Tests/UnitTests/EdgeHitProcessorTests.swift
```diff
@@ -46,17 +47,37 @@ class EdgeHitProcessorTests: XCTestCase {
...
+            guard let identityMapData = """
+                {
+                  "ECID" : [
+                    {
+                      "authenticationState" : "ambiguous",
+                      "id" : "test-mcid",
+                      "primary" : false
+                    }
+                  ]
+                }
```
kevinlind, line 68: The IdentityEdge shared state contains a single key "identityMap" which then contains all the identifiers, so the data should look like:
```
{
  "identityMap" : {
    "ECID" : [
      {
        "authenticationState" : "ambiguous",
        "id" : "test-mcid",
        "primary" : false
      }
    ]
  }
}
```

Tests/UnitTests/EdgeRequestTests.swift
```diff
@@ -24,9 +24,22 @@ class EdgeRequestTests: XCTestCase {
...
+        guard let identityMapData = """
+            {
+              "email" : [
+                {
+                  "authenticationState" : "ambiguous",
+                  "id" : "example@adobe.com",
+                  "primary" : false
+                }
+              ]
+            }
```
kevinlind, line 36: Same comment as above, the IdentityEdge shared state has a single property "identityMap" which contains all the identities. The test data should look like:
```
{
  "identityMap" : {
    "email" : [
      {
        "authenticationState" : "ambiguous",
        "id" : "example@adobe.com",
        "primary" : false
      }
    ]
  }
}
```

Tests/UnitTests/RequestBuilderTests.swift
```diff
@@ -31,7 +31,21 @@ class RequestBuilderTests: XCTestCase {
     func testGetRequestPayload_allParameters_verifyMetadata() {
         let request = RequestBuilder()
         request.enableResponseStreaming(recordSeparator: "A", lineFeed: "B")
-        request.experienceCloudId = "ecid"
+        guard let identityMapData = """
```
kevinlind, line 34: Same comment in this file about adding "identityMap" to the test shared states.

Tests/UnitTests/EdgeRequestTests.swift
nporter-adbe: In this test we populate `RequestContextData.identityMap` with the JSON shown in the PR. If we were to use the JSON suggested we would get a strange looking request when we flatten it:
```
XCTAssertEqual failed: ("{
    "xdm.identityMap.email[0].id" = "example@adobe.com";
}") is not equal to ("{
    "xdm.identityMap.identityMap.email[0].id" = "example@adobe.com";
}")
```

The `RequestContextData` type already has a key property `identityMap`, we then populate it with what would be the contents of the `identityMap` object. So in this test, I think this is the valid thing to do.

Tests/UnitTests/RequestBuilderTests.swift
nporter-adbe: Same as my above comment, we are populating the "identityMap" field on the request.

Tests/UnitTests/EdgeHitProcessorTests.swift
nporter-adbe: In this case, you're correct. I'll update it.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -74,12 +77,8 @@ class EdgeHitProcessor: HitProcessing {
         let requestBuilder = RequestBuilder()
         requestBuilder.enableResponseStreaming(recordSeparator: EdgeConstants.Defaults.RECORD_SEPARATOR,
                                                lineFeed: EdgeConstants.Defaults.LINE_FEED)
-
-        if let ecid = identityState[EdgeConstants.SharedState.Identity.ECID] as? String {
-            requestBuilder.experienceCloudId = ecid
-        } else {
-            // This is not expected to happen. Continue without ECID
-            Log.warning(label: LOG_TAG, "processHit - An unexpected error has occurred, ECID is nil.")
+        if let identityMap = identityState[EdgeConstants.SharedState.Identity.IDENTITY_MAP] as? [String: Any] {
```
nporter-adbe, line 80: Knowing the format of the IdentityMap is not optimal here. In a perfect situation, we could blindly attach the shared state, however, there is a bit of a limitation with `Encodable. We must know the keys we are going to encode before we encode them.

We should create a task for future work to do some fancy encoding so we can dynamically key our encoding container based on which ever XDM payloads are attached.

Tests/UnitTests/EdgeRequestTests.swift
nporter-adbe: Updated this. I got dynamic encoding to work.

Tests/UnitTests/RequestBuilderTests.swift
nporter-adbe: Updated this. I got dynamic encoding to work.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
nporter-adbe: Nevermind, I came up with a solution for not knowing the CodingKeys ahead of time. https://github.com/adobe/aepsdk-edge-ios/pull/165/commits/e7dfd3a3b52749bc74ed6b55730883cfb2361065#diff-1c5046ba14094114b426fa3ad5722b3901cb88b555ffdfdef0beb37683d51f1bR32

Sources/EdgeNetworkHandlers/RequestBuilder.swift
```diff
@@ -80,13 +76,12 @@ class RequestBuilder {
         guard event.data != nil,
               let consents = event.data?[EdgeConstants.EventDataKeys.CONSENTS] as? [String: Any] else { return nil }
 
-        // set ECID if available
-        var identityMap = IdentityMap()
-        if let ecid = experienceCloudId {
-            identityMap.addItem(namespace: EdgeConstants.JsonKeys.ECID, id: ecid)
+        // set IdentityMap if available
+        var identityMap = [String: AnyCodable]()
+        if let identityMapDict = xdmPayloads.first(where: {$0.keys.first == EdgeConstants.SharedState.Identity.IDENTITY_MAP}) {
```
nporter-adbe, line 81: This is a bit of an edge case, realistically we shouldn't know anything about the structure of the IdentityMap. We could do this by doing some custom encoding on the consent payload, but it seems a bit overkill given we aren't going to be attaching any additional XDM payloads on a consent request.

I think it's a fair compromise to say that Edge knows that name of "identityMap" so it can append to the Consent hit.

Sources/EdgeNetworkHandlers/RequestContextData.swift
```diff
@@ -10,10 +10,34 @@
...
 //
 
+import AEPServices
 import Foundation
 
 /// Property that holds the global XDM context data within an `EdgeRequest` object.
 /// It is contained within the `EdgeRequest` request property.
 struct RequestContextData: Encodable {
-    let identityMap: IdentityMap?
+    var xdmPayloads: [[String: AnyCodable]] = []
```
emdobrin, line 19: I don't think we need the RequestContextData struct anymore, we can just use 
```
/// struct EdgeRequest
let xdm: [String: AnyCodable]?
```

Tests/UnitTests/EdgeHitProcessorTests.swift
```diff
@@ -51,17 +52,39 @@ class EdgeHitProcessorTests: XCTestCase {
...
 
+    private func resolveXDMSharedState(extensionName: String, event: Event?) -> SharedStateResult? {
+        if extensionName == IDENTITY_SHARED_STATE {
+            guard let identityMapData = """
+                {
+                  "identityMap" : {
+                    "ECID" : [
+                      {
+                        "authenticationState" : "ambiguous",
+                        "id" : "test-mcid",
```
emdobrin, line 70: test-ecid

Tests/UnitTests/RequestBuilderTests.swift
```diff
@@ -43,17 +60,38 @@ class RequestBuilderTests: XCTestCase {
...
+                }
+              ]
+            }
+        }
+        """.data(using: .utf8) else {
+            XCTFail("Failed to convert json string to data")
+            return
+        }
+        let identityMap = try? JSONSerialization.jsonObject(with: identityMapData, options: []) as? [String: Any]
+        request.xdmPayloads += [AnyCodable.from(dictionary: identityMap)!]
```
emdobrin, line 94: can we extract this in a helper or property and reuse it for all tests?

Sources/EdgeNetworkHandlers/RequestContextData.swift
nporter-adbe: I'd prefer to keep the struct as-is since it has custom encoding. It makes testing easy, and we don't need to have a custom encoding for the entire request.

Sources/EdgeNetworkHandlers/RequestBuilder.swift
```diff
@@ -23,8 +23,8 @@ class RequestBuilder {
     /// Control character used at the end of each response fragment. Response streaming is enabled when both `recordSeparator` and `lineFeed` are non nil.
     private var lineFeed: String?
 
-    /// The Experience Cloud ID to be sent with this request
-    var experienceCloudId: String?
+    /// XDM payloads to be attached to the request
+    var xdmPayloads: [[String: AnyCodable]] = []
```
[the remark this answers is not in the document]
nporter-adbe: I made the change as suggested locally, but I actually find the original name of `xdmPayloads` to be more concise and self-explanatory.

Example:
`requestBuilder.xdmPayloads = [AnyCodable.from(dictionary: identityState) ?? [:]]`
vs
`requestBuilder.requestXdmPayload = [AnyCodable.from(dictionary: identityState) ?? [:]]`

1. In many cases `request` in the variable name for the XDM payloads is redundant given it's attached to the request builder.
2. The original name indicates its plural, indicating it could contain multiple XDM payloads.

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
kevinlind: IdentityEdge dev no longer sets a standard shared state. Can you update the Pods and change the test expectations?

kevinlind (approved):

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -74,12 +77,7 @@ class EdgeHitProcessor: HitProcessing {
 
         // Build Request object
         let requestBuilder = RequestBuilder()
-        if let ecid = identityState[EdgeConstants.SharedState.Identity.ECID] as? String {
-            requestBuilder.experienceCloudId = ecid
-        } else {
-            // This is not expected to happen. Continue without ECID
-            Log.warning(label: LOG_TAG, "processHit - An unexpected error has occurred, ECID is nil.")
-        }
+        requestBuilder.xdmPayloads = [AnyCodable.from(dictionary: identityState) ?? [:]]
```
emdobrin, line 80: We might need to extract the identityMap here as well. If Identity XDM Shared States gets updated with new fields in the future and we shouldn't include those in the Edge requests, this will break the network requests for old implementations of IdentityEdge.

Sources/EdgeNetworkHandlers/RequestBuilder.swift
emdobrin: I think it is correct to say we want to attach the identityMap to this request

Sources/EdgeNetworkHandlers/RequestContextData.swift
emdobrin: But with your changes this struct is no longer used for other custom fields, the EdgeRequest doesn't have ContextData, but rather meta, xdm (which represents the xdm payloads applied for the entire request) and events.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
nporter-adbe: If this is the case, then this breaks the contract that we have laid out in previous conversations of the Edge extension being able to blindly attach the IdentityMap.

If that is required we can change later, but we shouldn't be pulling out the IdentityMap then grooming it before the network request is sent out. I thought the point was that the map would be in the correct format and we would not need to make any modifications to it before sending it.

emdobrin: I am refering to the case where com.adobe.identityedge XDM shared state becomes:

nporter-adbe: I find it a bit odd that we must parse the Identity shared state in the Edge extension, but I've updated the PR to only include the IdentityMap.
https://github.com/adobe/aepsdk-edge-ios/pull/165/commits/ecd2dcbb9872357a2ee3e591262d97b1ee203a68

emdobrin (approved):

PR #166, opened 2021-03-02
[AMSDK-11130] Collect consent

emdobrin:
## Description

- Adds Edge + Consent integration and updates the hit queue based on current consent status as follows: 
    - When collect consent is yes, the hitQueue is resumed and data is being sent to the Edge network.
    - When collect consent is no, the previously queued hits are cleared from persistence and no new experience events are queued anymore. Consent update events are still queued and sent to the Edge network.
    - When collect consent is pending, the queue is suspended waiting for a new consent update to yes/no.
    - When Consent extension is not registered, the default collect consent status is yes.
- Drops the integration with core privacy status in favor of consent settings.
- Tests for consent update requests and experience events handling based on current collect consent status.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.


Tests/FunctionalTests/EdgePublicAPITests.swift
```diff
@@ -0,0 +1,84 @@
...
+// the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+// OF ANY KIND, either express or implied. See the License for the specific language
+// governing permissions and limitations under the License.
+//
+
+@testable import AEPCore
+@testable import AEPEdge
+import XCTest
+
+class EdgePublicAPITests: XCTestCase {
```
emdobrin, line 17: Replaces EdgeTests

Tests/TestUtils/MockDataQueue.swift
```diff
@@ -0,0 +1,60 @@
...
+// Unless required by applicable law or agreed to in writing, software distributed under
+// the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+// OF ANY KIND, either express or implied. See the License for the specific language
+// governing permissions and limitations under the License.
+//
+
+@testable import AEPServices
+import Foundation
+
+/// MockDataQueue - see also AEPServices/Mocks
```
emdobrin, line 16: MockDataQueue, MockHitProcessor and TestableExtensionRuntime are copied from core.

Tests/FunctionalTests/Edge+ConsentTests.swift
```diff
@@ -0,0 +1,276 @@
...
+        updateCollectConsent(status: "u")
+        updateCollectConsent(status: "some value")
+
+        // verify
+        let interactRequests = self.getNetworkRequestsWith(url: FunctionalTestConst.EX_EDGE_INTERACT_URL_STR, httpMethod: HttpMethod.post, timeout: 1)
+        XCTAssertEqual(0, interactRequests.count)
+        let consentRequests = self.getNetworkRequestsWith(url: FunctionalTestConst.EX_EDGE_CONSENT_URL_STR, httpMethod: HttpMethod.post, timeout: 1)
+        XCTAssertEqual(0, consentRequests.count)
+    }
+
```
PravinPK, line 254: good to have a few more tests here.. with Consents obtained from defaultConfig only and no Consent API's are called

I guess we for those test we should wait for that Bootup Consent PR to be merged.

Sources/EdgeHitQueueing+Consent.swift
```diff
@@ -0,0 +1,32 @@
...
+extension HitQueuing {
+    /// Based on `status` determines if we should continue processing hits or if we should suspend processing and clear hits
+    /// - Parameter status: the current collect consent status status
+    func handleCollectConsentChange(status: ConsentStatus) {
+        switch status {
+        case .yes:
+            beginProcessing()
+        case .no:
+            clear()
+            beginProcessing()
```
PravinPK, line 25: do we need beginProcessing() here ?

Sources/EdgeHitQueueing+Consent.swift
```diff
@@ -0,0 +1,32 @@
...
+// OF ANY KIND, either express or implied. See the License for the specific language
+// governing permissions and limitations under the License.
+//
+
+import AEPServices
+import Foundation
+
+extension HitQueuing {
+    /// Based on `status` determines if we should continue processing hits or if we should suspend processing and clear hits
+    /// - Parameter status: the current collect consent status status
```
PravinPK, line 18: consent status status.... :)

Sources/Edge.swift
```diff
@@ -123,93 +116,63 @@ public class Edge: NSObject, Extension {
...
+            Log.debug(label: LOG_TAG, "Ignoring event with id \(event.id.uuidString) due to collect consent setting (n) .")
             return true
         }
 
         return false
     }
 
+    /// Determines if `Edge` is ready to handle events, if the bootup can be executed successfully
+    /// - Parameter event: An `Event`
+    /// - Returns: True if we can process events, false otherwise
```
PravinPK, line 139: trivial 
-Returns: `t`rue

PravinPK (approved):
Neat. The logic looks good to me

Sources/EdgeHitQueueing+Consent.swift
emdobrin: Yes, this is required for us to keep queuing and processing consent update events while collect consent is no, while ignoring experience events.

Tests/FunctionalTests/Edge+ConsentTests.swift
```diff
@@ -0,0 +1,276 @@
...
+        "              \"maxAge\": 7200" +
+        "            }" +
+        "          ]," +
+        "          \"type\": \"state:store\"" +
+        "        }" +
+        "      ]" +
+        "    }\n"
+
+    public class override func setUp() {
+        super.setUp()
```
nporter-adbe, line 51: Do we need to call `super.setUp()`? I don't recall seeing these before.

Sources/Edge.swift
```diff
@@ -19,8 +19,7 @@ public class Edge: NSObject, Extension {
     private let LOG_TAG = "Edge" // Tag for logging
     private var networkService: EdgeNetworkService = EdgeNetworkService()
     private var networkResponseHandler: NetworkResponseHandler?
-    private var hitQueue: HitQueuing?
-    private var currentPrivacyStatus: PrivacyStatus?
+    internal var state: EdgeState?
```
nporter-adbe, line 22: What's the reason for this being internal? Testing?

Sources/Edge.swift
```diff
@@ -87,32 +89,23 @@ public class Edge: NSObject, Extension {
...
-            if privacyStatus == .optedOut {
-                let storeResponsePayloadManager = StoreResponsePayloadManager(EdgeConstants.DataStoreKeys.STORE_NAME)
-                storeResponsePayloadManager.deleteAllStorePayloads()
-                Log.debug(label: LOG_TAG, "Device has opted-out of tracking. Clearing the Edge queue.")
-            }
-        }
+    /// Handles the `EventType.consent` -`EventSource.responseContent` event for the collect consent change
+    /// - Parameter event: the consent preferences response event
+    func handleConsentPreferencesUpdate(_ event: Event) {
+        guard let data = event.data else { return }
```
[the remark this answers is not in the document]
nporter-adbe: Also would probably be good to add a log that we are dropping the event as data is nil/empty.

Sources/EdgeConsentStatus.swift
```diff
@@ -12,7 +12,37 @@
...
+
+    /// Initializes the appropriate `ConsentStatus` enum for the given `rawValue`
+    /// - Parameter rawValue: a `RawValue` representation of a `ConsentStatus` enum, default is pending
+    public init(rawValue: RawValue) {
+        switch rawValue {
+        case "y":
+            self = .yes
+        case "n":
+            self = .no
+        default:
```
nporter-adbe, line 30: The default case is a bit confusing for me, as we have two default values in our constants based on specific scenarios. I wonder if it's the correct thing to do to default to pending here.

What if we made it a failable initializer and returned nil instead of providing a default value as it can vary based on the situation?

`init?(rawValue: Self.RawValue)`

Sources/EdgeState.swift
```diff
@@ -0,0 +1,64 @@
...
+    func bootupIfNeeded(event: Event, getXDMSharedState: @escaping (String, Event?) -> SharedStateResult?) {
+        guard !hasBooted else { return }
+
+        // check if consent extension is registered
+        let consentSharedState = getXDMSharedState(EdgeConstants.SharedState.Consent.SHARED_OWNER_NAME, nil)
+        if consentSharedState == nil {
+            Log.warning(label: LOG_TAG, "Consent extension is not registered yet, using default collect status (yes)")
+            currentCollectConsent = EdgeConstants.Defaults.COLLECT_CONSENT_YES
+        } else {
+            currentCollectConsent = ConsentStatus.getCollectConsentOrDefault(eventData: consentSharedState?.value ?? [:])
```
nporter-adbe, line 46: I find it a bit odd that we check if `consentSharedState` is nil but we are still forced to do nil coalescing with `consentSharedState?.value ?? [:]`

Maybe we can just write this as an if-let to unwrap the optional.

```swift
if let consentSharedState = getXDMSharedState(EdgeConstants.SharedState.Consent.SHARED_OWNER_NAME, nil) {
    // ...
} else {
   // handle nil case
}
```

Sources/Edge.swift
```diff
@@ -34,29 +33,32 @@ public class Edge: NSObject, Extension {
         super.init()
 
         // set default on init for register/unregister use-case
-        currentPrivacyStatus = EdgeConstants.DEFAULT_PRIVACY_STATUS
-        networkResponseHandler = NetworkResponseHandler(getPrivacyStatus: getPrivacyStatus)
-        setupHitQueue()
+        networkResponseHandler = NetworkResponseHandler()
+        if let hitQueue = setupHitQueue() {
```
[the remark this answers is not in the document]
emdobrin: We do log in setupHitQueue.

Tests/FunctionalTests/Edge+ConsentTests.swift
emdobrin: yes, we are overriding both class & test setUp functions in FunctionalTestBase, so this will call the base class setUp first:
```
demo - class setUp FunctionalTestBase
demo - class setUp my test
demo - setUp FunctionalTestBase
demo - setUp my test
```

Tests/FunctionalTests/Edge+ConsentTests.swift
emdobrin: There are some more tests for this scenario in EdgeExtensionTests -> testBootup with different Consent shared states (see testBootup_whenNoConsentSharedState_usesDefaultYes)

Sources/EdgeConsentStatus.swift
emdobrin: Don't think it will help much because I am already unwrapping the string before calling CollectConsent(rawValue:) - see getCollectConsentOrDefault. Changed to return the Defaults.COLLECT_CONSENT_PENDING.

The only case where we use default yes is on bootup when Consent extension is not registered.

Sources/Edge.swift
emdobrin: overwriting it in EdgeExtensionTests

nporter-adbe (approved):

Sources/EdgeState.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// or the default value if this extension is not registered
+    ///
+    /// - Parameters:
+    ///   - event: The `Event` triggering the bootup
+    ///   - getXDMSharedState: used to fetch the Consent data
+    func bootupIfNeeded(event: Event, getXDMSharedState: @escaping (String, Event?) -> SharedStateResult?) {
+        guard !hasBooted else { return }
+
+        // check if consent extension is registered
+        if let consentSharedState = getXDMSharedState(EdgeConstants.SharedState.Consent.SHARED_OWNER_NAME, nil) {
```
nporter-adbe, line 41: I think we should be reading the EventHub shared state to determine if the consent extension is registered. I think this approach of checking if the shared state exists to determine if an extension is registered can lead to false negatives.

We've used this method in our legacy codebase before, but I think we've found it tricky in some situations. Just because shared state doesn't exist for an extension doesn't mean it doesn't exist.

emdobrin: I looked at using either the shared state or MobileCore.getRegisteredExtensions API but found that it only includes the friendly name and version of each extension, whereas other extensions should rely on is the extension name (aka shared state name) as the friendly name can anytime change. I discussed a bit about this with @shalehaha and we are considering it, but are still evaluating the implications for Assurance UI before making any changes.

I am willing to change this logic to check for hub shared state/getRegisteredExtensions as soon as the extension name is included in there.

Issue #84, opened 2020-10-21
Configurable Edge endpoint for development testing


emdobrin:
Need to provide a way for developers to test with Edge pre-prod/integration. 

Proposed solution: expose a new configuration key for the Experience Edge testing environment only for development, that can be set using MobileCore.updateConfig API.
Edge extension reads this key from config and does the mapping to the correct endpoint. By default, if this key is not present or it is not one of the expected values, it falls back to prod.

This key will not be exposed in Launch UI.

e.g. "edge.environment" = String in ("prod", "pre-prod", "int")

emdobrin:
@kevinlind @pfransenadb thoughts on this? This is pretty frequently used when testing new features in Experience Edge pre-prod/int.

pfransenadb:
@emdobrin I'm fine with adding this key for the purpose you describe.  Makes sense to me.

kevinlind:
@emdobrin I like this idea too. It will be useful in cases where a custom is asking for a fix/change in Edge and needs to verify it before Edge pushes it to production. For clarification, when you say "only for development", are you meaning the config key is only used when the Launch "build.environment" is "dev"?

emdobrin:
@kevinlind  I mean only for development testing, when using MobileCore.updateConfig and not supported in the Edge Launch UI.

PR #168, opened 2021-03-17
Build then persist Edge hits

nporter-adbe:
## Description

Instead of just persisting the Event we now build the Edge hit then persist the entire hit.

## Types of changes

## Checklist:

### DIFF
--- a/Sources/Edge.swift
+++ b/Sources/Edge.swift
@@ -84,13 +84,51 @@ public class Edge: NSObject, Extension {
             return
         }
 
-        guard let eventData = try? JSONEncoder().encode(event) else {
-            Log.debug(label: LOG_TAG, "handleExperienceEventRequest - Failed to encode event with id: '\(event.id.uuidString)'.")
+        // fetch config shared state, this should be resolved based on readyForEvent check
+        guard let configId = getEdgeConfigId(event: event) else {
+            Log.debug(label: LOG_TAG, "Unable to read Edge config id, dropping event with id: \(event.id.uuidString)")
+            return // drop current event
+        }
+
+        // get IdentityMap from Identity shared state, this should be resolved based on readyForEvent check
+        guard let identityState =
+                getXDMSharedState(extensionName: EdgeConstants.SharedState.Identity.STATE_OWNER_NAME,
+                                  event: event)?.value else {
+            Log.warning(label: LOG_TAG,
+                        "handleExperienceEventRequest - Unable to process the event '\(event.id.uuidString)', " +
+                            "Identity shared state is nil.")
+            return // drop current event
+        }
+
+        // Build Request object
+        let requestBuilder = RequestBuilder()
+        // attach identity map
+        requestBuilder.xdmPayloads[EdgeConstants.SharedState.Identity.IDENTITY_MAP] =
+            AnyCodable(identityState[EdgeConstants.SharedState.Identity.IDENTITY_MAP])
+
+        requestBuilder.enableResponseStreaming(recordSeparator: EdgeConstants.Defaults.RECORD_SEPARATOR,
+                                               lineFeed: EdgeConstants.Defaults.LINE_FEED)
+
+        // Build and send the network request to Experience Edge
+        let listOfEvents: [Event] = [event]
+        guard let requestPayload = requestBuilder.getPayloadWithExperienceEvents(listOfEvents) else {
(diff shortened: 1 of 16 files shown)


nporter-adbe:
> Merging #168 (a6859f4) into feature/consent (42c7db2) will increase coverage by 0.00%.

🙃

emdobrin (changes requested):
@nporter-adbe thanks for putting this up. I have a few general request for this PR:

1. I think we should continue queuing events and not EdgeHits as this will allow us to support batching in the future (for example if multiple experience events have same store and identityMap, we can send one network request containing all).

2. We should queue the identityMap and current state store with the event so that the data is associated correctly, but we should not store the configuration and assurance information. The last two should be fetched on the fly when the request is being sent to the server, as the events can be processed in another session and config may be invalid at some point for any reason and it can block the entire queue.

Based on the above I am proposing the following changes: create a EdgeDataEntity class that holds the required information to be queued (event, state store, identityMap). At process time, build the Edge hit based on the event type/source and fetch config and request headers. Having a well defined EdgeDataEntity allows us to add new data in the future if needed for queued hits, without requiring a database migration.

nporter-adbe:
@emdobrin, I think the batching comment is a good point, but I have some concerns with the remaining feedback.

1. Config fetching:  The current implementation does not block the queue when the config is invalid. Instead, it just drops events which is consistent with the behavior in other extensions and platforms. If we have the scenario:
 
1.1: App is launched with an invalid config
1.2: Hits are queued with this invalid config
1.3 Config updated to a valid config
1.4 Hits are queued

It doesn't seem correct that we would assume that hits queued in 1.2 would automatically be associated with the new config in 1.3. It seems undesired to associate hits in 1.2 with a configuration that didn't exist when they were queued. Thus they should be dropped since they were queued when the SDK was in an invalid state.

2. Assurance integration ID: I still think that persisting the Assurance integration ID is the correct approach. Imagine the scenario where you queue a hit that is persisted in an app that does not contain Assurance, and that hit doesn't go out before the app is closed. Then you relaunch the app with Assurance included. I wouldn't expect a hit that was queued when Assurance wasn't present to contain an integration ID.

3. Database migration: I don't see how an `EdgeDataEntity` would have any database migration advantages over what is currently the `EdgeHit`. Both would be `Codable` types with well-defined data, and adding/removing properties would require migrating the type from persistence.

nporter-adbe:
Closing this, will open a new PR with your feedback, but still have some concerns as listed above.

emdobrin:
1. The configId might be worth storing in the data entity as  this shouldn't block the queue in any way because it is a query parameter in the request. But we need to think more about any edge-cases here: like config with org id1, state store with orgid 1, edge configid1, config updated with org id2, edge config2. 
Also, if we need to store other config keys in the future, for example server name or the edge endpoint (https://github.com/adobe/aepsdk-edge-ios/issues/84), we have to handle on the fly. These are keys that can be modified with remote configuration, at runtime, so if any hit gets stuck in the queue it requires app update to unblock it.

2. Assurance can be enabled at any point at runtime, and it should catch all the requests from the moment it was started, so I think we need to pull the integration id at processing time. If I reverse your example and I have an assurance session, queued the hits, closed the app, reopen app without assurance session, I am expecting not to see the assurance integration id in these calls. I should only see that when assurance is active at the moment of event processing.

3. I think the confusing part here is the name `EdgeHit`. That class was originally added for encapsulating the fields required for an Edge network request. The `EdgeHit` and the `EdgeDataEntity` have different fields based on how they are used. Here is a sample:
```
EdgeDataElement {
   let event: Event // only one event queued at a time
   let identityMap: [String: Any]
   let stateStore: [String: Any]?

   var newOptionalField: String? // new field in the future, preferable it should be optional so it doesn't break compatibility and it doesn't require db migration
   let newRequiredField: String // requires migration
}
```
```
EdgeHit {
    var configId: String 
    var requestId: String
    func getType() -> ExperienceEdgeRequestType
    func getPayload() -> String?
    func getStreamingSettings() -> Streaming?
}

ExperienceEventsEdgeHit {
   EdgeHit + EdgeRequest (list of events when batching collected from multiple EdgeDataElements)
}
```
That's why I think the db migration shouldn't be impacted by changes in the network request payload `EdgeHit` and subclasses, but it should be impacted by changes in the `EdgeDataEntity`, if it introduces required properties for processing.

PR #169, opened 2021-03-18
Handle reset identities event

nporter-adbe:
## Description

Instead of just storing the `Event` we now store the current identity shared state and the current store payloads. This will allow for easier implementation of handling the reset request event.

## Types of changes

## Checklist:

### DIFF
--- a/Sources/Edge.swift
+++ b/Sources/Edge.swift
@@ -110,7 +110,7 @@ public class Edge: NSObject, Extension {
                                         identityMap: AnyCodable.from(dictionary: identityState) ?? [:])
 
         guard let entityData = try? JSONEncoder().encode(edgeEntity) else {
-            Log.debug(label: LOG_TAG, "handleExperienceEventRequest - Failed to encode Edge data entity with id: '\(event.id.uuidString)'.")
+            Log.debug(label: LOG_TAG, "handleExperienceEventRequest - Failed to encode EdgeDataEntity with id: '\(event.id.uuidString)'.")
             return
         }
 
@@ -136,12 +136,12 @@ public class Edge: NSObject, Extension {
         let edgeEntity = EdgeDataEntity(event: event, identityMap: [:])
 
         guard let entityData = try? JSONEncoder().encode(edgeEntity) else {
-            Log.debug(label: LOG_TAG, "handleIdentitiesReset - Failed to encode Edge data entity with id: '\(event.id.uuidString)'.")
+            Log.debug(label: LOG_TAG, "handleIdentitiesReset - Failed to encode EdgeDataEntity with id: '\(event.id.uuidString)'.")
             return
         }
 
         let entity = DataEntity(uniqueIdentifier: event.id.uuidString, timestamp: event.timestamp, data: entityData)
-        networkResponseHandler?.lastResetDate = event.timestamp
+        networkResponseHandler?.setLastReset(date: event.timestamp)
         state?.hitQueue.queue(entity: entity)
     }
 
@@ -167,7 +167,7 @@ public class Edge: NSObject, Extension {
                                         identityMap: AnyCodable.from(dictionary: identityState) ?? [:])
 
         guard let entityData = try? JSONEncoder().encode(edgeEntity) else {
-            Log.debug(label: LOG_TAG, "handleConsentUpdate - Failed to encode Edge data entity with id: '\(event.id.uuidString)'.")
+            Log.debug(label: LOG_TAG, "handleConsentUpdate - Failed to encode EdgeDataEntity with id: '\(event.id.uuidString)'.")
             return
         }
 
--- a/Sources/Edge.swift
+++ b/Sources/Edge.swift
@@ -49,6 +49,9 @@ public class Edge: NSObject, Extension {
         registerListener(type: EventType.edge,
                          source: EventSource.updateConsent,
                          listener: handleConsentUpdate)
+        registerListener(type: EventType.genericIdentity,
+                         source: EventSource.requestReset,
+                         listener: handleIdentitiesReset)
     }
 
     public func onUnregistered() {
@@ -64,6 +67,15 @@ public class Edge: NSObject, Extension {
                                                           event: event)
             let identitySharedState = getXDMSharedState(extensionName: EdgeConstants.SharedState.Identity.STATE_OWNER_NAME,
                                                         event: event)
+
+            return configurationSharedState?.status == .set && identitySharedState?.status == .set
+        } else if event.isResetIdentitiesEvent {
+            let configurationSharedState = getSharedState(extensionName: EdgeConstants.SharedState.Configuration.STATE_OWNER_NAME,
+                                                          event: event)
+            // use barrier to wait for EdgeIdentity to handle the reset
+            let identitySharedState = getXDMSharedState(extensionName: EdgeConstants.SharedState.Identity.STATE_OWNER_NAME,
+                                                        event: event,
+                                                        barrier: true)
 
             return configurationSharedState?.status == .set && identitySharedState?.status == .set
         }
@@ -94,10 +106,8 @@ public class Edge: NSObject, Extension {
             return // drop current event
         }
 
-        let storeResponsePayloadManager = StoreResponsePayloadManager(EdgeConstants.DataStoreKeys.STORE_NAME)
         let edgeEntity = EdgeDataEntity(event: event,
-                                        identityMap: AnyCodable.from(dictionary: identityState) ?? [:],
-                                        storedPayloads: storeResponsePayloadManager.getActiveStores())
+                                        identityMap: AnyCodable.from(dictionary: identityState) ?? [:])
 
         guard let entityData = try? JSONEncoder().encode(edgeEntity) else {
             Log.debug(label: LOG_TAG, "handleExperienceEventRequest - Failed to encode Edge data entity with id: '\(event.id.uuidString)'.")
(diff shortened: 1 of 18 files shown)


kevinlind (commented):
This looks good. However, I have questions regarding how the response payload is updated from the server response. This implementation would mean that any queued hit's stored payload would not get updated by the server response of earlier hits.  That's different than what we do currently when the stored payloads are retrieved right before the hit is sent.  

Also, if the hit queue is stored for a long time (app closed for a while then relaunched), there's the possibility of a stored payload timing out. Those timed out payload elements are currently removed before sending a hit, but with this change they are not. Will that cause issues with the server?

nporter-adbe:
@kevinlind  Correct, the stored payload of queued hits are not able to be updated. I'm on the fence about which approach is better. On one end I see that we should get a snapshot of the current cookies when the hit was queued and use them, but I can also see that always getting the latest cookies may be nice to have. @emdobrin thoughts?

That's a good point regarding expired payloads... I would hope the server could handle it, but I made a small update to filter out expired payloads just in case. I should probably add a test for this edge case, but let me know your thoughts.
https://github.com/adobe/aepsdk-edge-ios/pull/169/commits/77f7b7bdd2d2807f844a5632f8ac9f7b8ce9f4a3

emdobrin:
I think you are right, we cannot easily update the persisted state store if they change in between hits, as you have to update certain events, but not all. One other option is to have a combination of the two approaches we discussed:
- keep the state store on demand (don't persist it)
- persist the reset event
- when processing the reset event clear store and save its timestamp
- if state:store response comes for a request event with timestamp < than reset event timestamp, ignore it (also don't dispatch this state:store for others to consume)

In other words:
1. If state:store is updated while processing events, the next hits will contain the updated information, as soon as the response is processed client-side.
2. If state:store is updated async after a reset event was processed, it doesn't overwrite the cleared store.

nporter-adbe:
@kevinlind @emdobrin I pushed up some changes to handle the reset event in this PR as well and took into account some of the feedback of not persisting the state store payloads.

The PR isn't in final form and needs some tests added, but let me know what you think about this approach.

Sources/Edge.swift
```diff
@@ -65,6 +68,15 @@ public class Edge: NSObject, Extension {
             let identitySharedState = getXDMSharedState(extensionName: EdgeConstants.SharedState.Identity.STATE_OWNER_NAME,
                                                         event: event)
 
+            return configurationSharedState?.status == .set && identitySharedState?.status == .set
+        } else if event.isResetIdentitiesEvent {
+            let configurationSharedState = getSharedState(extensionName: EdgeConstants.SharedState.Configuration.STATE_OWNER_NAME,
+                                                          event: event)
+            // use barrier to wait for EdgeIdentity to handle the reset
```
emdobrin, line 75: so are we still waiting for the reset request and not for the resetComplete?

Sources/Edge.swift
```diff
@@ -84,13 +96,26 @@ public class Edge: NSObject, Extension {
...
+                        "handleExperienceEventRequest - Unable to process the event '\(event.id.uuidString)', " +
+                            "Identity shared state is nil.")
+            return // drop current event
+        }
+
+        let edgeEntity = EdgeDataEntity(event: event,
+                                        identityMap: AnyCodable.from(dictionary: identityState) ?? [:])
+
+        guard let entityData = try? JSONEncoder().encode(edgeEntity) else {
+            Log.debug(label: LOG_TAG, "handleExperienceEventRequest - Failed to encode Edge data entity with id: '\(event.id.uuidString)'.")
```
emdobrin, line 113: Failed to encode EdgeDataEntity for event with id..

Sources/Edge.swift
```diff
@@ -105,6 +130,21 @@ public class Edge: NSObject, Extension {
         state?.updateCurrentConsent(status: ConsentStatus.getCollectConsentOrDefault(eventData: data))
     }
 
+    /// Handles the generic identities reset event
+    /// - Parameter event: an `Event`
+    func handleIdentitiesReset(_ event: Event) {
+        let edgeEntity = EdgeDataEntity(event: event, identityMap: [:])
+
+        guard let entityData = try? JSONEncoder().encode(edgeEntity) else {
+            Log.debug(label: LOG_TAG, "handleIdentitiesReset - Failed to encode Edge data entity with id: '\(event.id.uuidString)'.")
```
emdobrin, line 139: Failed to encode EdgeDataEntity for event with id..

Sources/Edge.swift
```diff
@@ -113,12 +153,25 @@ public class Edge: NSObject, Extension {
...
+                        "handleConsentUpdate - Unable to process the event '\(event.id.uuidString)', " +
+                            "Identity shared state is nil.")
+            return // drop current event
+        }
+
+        let edgeEntity = EdgeDataEntity(event: event,
+                                        identityMap: AnyCodable.from(dictionary: identityState) ?? [:])
+
+        guard let entityData = try? JSONEncoder().encode(edgeEntity) else {
+            Log.debug(label: LOG_TAG, "handleConsentUpdate - Failed to encode Edge data entity with id: '\(event.id.uuidString)'.")
```
emdobrin, line 170: Failed to encode EdgeDataEntity for event with id..

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -43,15 +43,15 @@ class EdgeHitProcessor: HitProcessing {
     }
 
     func processHit(entity: DataEntity, completion: @escaping (Bool) -> Void) {
-        guard let data = entity.data, let event = try? JSONDecoder().decode(Event.self, from: data),
-              let eventData = event.data, !eventData.isEmpty
+        guard let data = entity.data, let edgeEntity = try? JSONDecoder().decode(EdgeDataEntity.self, from: data)
         else {
             // can't convert data to hit, unrecoverable error, move to next hit
-            Log.debug(label: LOG_TAG, "processHit - Failed to decode edge hit with id '\(entity.uniqueIdentifier)' or empty data.")
+            Log.debug(label: LOG_TAG, "processHit - Failed to decode edge hit with id '\(entity.uniqueIdentifier)'.")
```
emdobrin, line 49: Failed to decode EdgeDataEntity with id ..

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -33,13 +36,12 @@ class NetworkResponseHandler {
...
         guard !requestId.isEmpty, !batchedEvents.isEmpty else { return }
 
-        let eventIds = batchedEvents.map { $0.id.uuidString }
         serialQueue.sync {
             if self.sentEventsWaitingResponse[requestId] != nil {
                 Log.warning(label: self.LOG_TAG, "Name collision for requestId \(requestId), events list is overwritten.")
             }
 
-            self.sentEventsWaitingResponse[requestId] = eventIds
+            self.sentEventsWaitingResponse[requestId] = batchedEvents
```
emdobrin, line 44: we might need to optimize this and only keep the event id and timestamp. if there are many events with large payload being processed at a time, this can bump the memory usage quickly.

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -66,13 +68,19 @@ class NetworkResponseHandler {
     ///   - requestId: request id associated with current response
     func processResponseOnSuccess(jsonResponse: String, requestId: String) {
         guard let data = jsonResponse.data(using: .utf8) else { return }
+        var ignoreStorePayloads = false
+        if let firstEvent = sentEventsWaitingResponse[requestId]?.first, let lastResetDate = lastResetDate {
```
emdobrin, line 72: hmm.. this is an interesting one.. we should probably add a comment when batching saying that `multiple events cannot be batched if there was a reset in between them`

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -66,13 +68,19 @@ class NetworkResponseHandler {
...
+            ignoreStorePayloads = firstEvent.timestamp < lastResetDate
+        }
```
emdobrin, line 74: let's extract in a new func `ignoreStorePayloads` or similar

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -144,7 +155,7 @@ class NetworkResponseHandler {
         guard let requestEventIdsList = getWaitingEvents(requestId: requestId) else { return nil }
 
         // Note: ExEdge does not return eventIndex when there is only one event in the request.
-        // The event handles and errrors are associated to that request event, so defaulting to 0 here.
+        // The event handles and errors are associated to that request event, so defaulting to 0 here.
```
emdobrin, line 158: roar 🐯

emdobrin (commented):
can you also add some integration tests involving experience events with resetIdentitities in between?

Sources/Edge.swift
nporter-adbe: Yeah, we could listen for the event coming from edge identity notifying it has reset or use the barrier flag to wait until we know its done processing

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
nporter-adbe: changed this to store tuples of `(uuid: String, date: Date)`

Sources/EdgeNetworkHandlers/Atomic.swift
```diff
@@ -0,0 +1,38 @@
...
+// Unless required by applicable law or agreed to in writing, software distributed under
+// the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+// OF ANY KIND, either express or implied. See the License for the specific language
+// governing permissions and limitations under the License.
+//
+
+import Foundation
+
+/// Represents a thread-safe type for read and write operations
+final class Atomic<A> {
```
nporter-adbe, line 16: We probably need to add some tests for this type, but let's add these later.

Tests/UnitTests/AtomicTests.swift
```diff
@@ -0,0 +1,31 @@
...
+
+@testable import AEPEdge
+import Foundation
+import XCTest
+
+class AtomicTests: XCTestCase {
+    private let queue = DispatchQueue(label: "test queue",
+                                      attributes: .concurrent)
+
+    func testAtomicWrite() {
```
nporter-adbe, line 21: We may want to add more tests in the future, but for now I think this is sufficient.

kevinlind (approved):
Looks good to me. Thanks for all the updates.

Sources/EdgeNetworkHandlers/Atomic.swift
```diff
@@ -0,0 +1,38 @@
...
+    private let queue = DispatchQueue(label: "Atomic type serial queue")
```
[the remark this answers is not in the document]
nporter-adbe: They will all be different queues, the label is only used for debugging. I don't think it really makes sense to need to specify the queue label when creating an `Atomic`, so I used a general label in all cases. `"Atomic type serial queue"`.

PR #171, opened 2021-03-23
Renaming and update dependencies

emdobrin:


PravinPK (approved):

kevinlind (commented):
Looks good. There's a separate Podfile for the commerce demo app which needs to be updated as well.

kevinlind (approved):

PR #189, opened 2021-04-15
[AMSDK-11410] Add support for configurable testing endpoints

nporter-adbe:
## Description

Implements: https://github.com/adobe/aepsdk-edge-ios/issues/84

## Types of changes

## Checklist:

### DIFF
--- a/Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
+++ b/Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
@@ -174,8 +174,8 @@ class EdgeHitProcessor: HitProcessing {
             return nil
         }
 
-        let edgeEndpointStr = configSharedState[EdgeConstants.SharedState.Configuration.EDGE_ENVIRONMENT] as? String ?? EdgeEndpoint.production.rawValue
-        let edgeEndpoint = EdgeEndpoint(rawValue: edgeEndpointStr.lowercased()) ?? EdgeEndpoint.production
+        let edgeEndpointStr = configSharedState[EdgeConstants.SharedState.Configuration.EDGE_ENVIRONMENT] as? String
+        let edgeEndpoint = EdgeEndpoint(optionalRawValue: edgeEndpointStr?.lowercased())
 
         guard let configId =
                 configSharedState[EdgeConstants.SharedState.Configuration.CONFIG_ID] as? String,
--- a/Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
+++ b/Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
@@ -57,7 +57,7 @@ class EdgeHitProcessor: HitProcessing {
         }
 
         // fetch config shared state, this should be resolved based on readyForEvent check
-        guard let configId = getEdgeConfigId(event: event) else {
+        guard let (configId, edgeEndpoint) = getEdgeConfig(event: event) else {
             completion(true)
             return // drop current event
         }
@@ -72,6 +72,7 @@ class EdgeHitProcessor: HitProcessing {
         // Enable response streaming for all events
         requestBuilder.enableResponseStreaming(recordSeparator: EdgeConstants.Defaults.RECORD_SEPARATOR,
                                                lineFeed: EdgeConstants.Defaults.LINE_FEED)
+
         if event.isExperienceEvent {
             guard let eventData = event.data, !eventData.isEmpty else {
                 Log.debug(label: EdgeConstants.LOG_TAG, "\(SELF_TAG) - Failed to process Experience event, data was nil or empty")
@@ -89,7 +90,7 @@ class EdgeHitProcessor: HitProcessing {
                 return
             }
 
-            let edgeHit = ExperienceEventsEdgeHit(configId: configId, request: requestPayload)
+            let edgeHit = ExperienceEventsEdgeHit(edgeEndpoint: edgeEndpoint, configId: configId, request: requestPayload)
             // NOTE: the order of these events needs to be maintained as they were sent in the network request
             // otherwise the response callback cannot be matched
             networkResponseHandler.addWaitingEvents(requestId: edgeHit.requestId,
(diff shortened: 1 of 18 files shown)


kevinlind (approved):
Looks good.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -172,6 +174,9 @@ class EdgeHitProcessor: HitProcessing {
             return nil
         }
 
+        let edgeEndpointStr = configSharedState[EdgeConstants.SharedState.Configuration.EDGE_ENVIRONMENT] as? String ?? EdgeEndpoint.production.rawValue
+        let edgeEndpoint = EdgeEndpoint(rawValue: edgeEndpointStr.lowercased()) ?? EdgeEndpoint.production
```
emdobrin, line 178: I would prefer to have a default in EdgeEndpoint enum so we don't need to handle this case here, and we can pass nil, empty strings, random strings without worry.

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
```diff
@@ -21,7 +21,9 @@ import XCTest
 class AEPEdgeFunctionalTests: FunctionalTestBase {
     private let event1 = Event(name: "e1", type: "eventType", source: "eventSource", data: nil)
     private let event2 = Event(name: "e2", type: "eventType", source: "eventSource", data: nil)
-    private let exEdgeInteractUrl = URL(string: FunctionalTestConst.EX_EDGE_INTERACT_URL_STR)! // swiftlint:disable:this force_unwrapping
+    private let exEdgeInteractProdUrl = URL(string: FunctionalTestConst.EX_EDGE_INTERACT_PROD_URL_STR)! // swiftlint:disable:this force_unwrapping
+    private let exEdgeInteractPreProdUrl = URL(string: FunctionalTestConst.EX_EDGE_INTERACT_PRE_PROD_URL_STR)! // swiftlint:disable:this force_unwrapping
+    private let exEdgeInteractIntegrationUrl = URL(string: FunctionalTestConst.EX_EDGE_INTERACT_INTEGRATION_URL_STR)! // swiftlint:disable:this force_unwrapping
```
emdobrin, line 26: Should we disable this rule in `Tests/TestUtils/.swiftlint.yml` ?

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
```diff
@@ -924,6 +926,159 @@ class AEPEdgeFunctionalTests: FunctionalTestBase {
...
+                                                                error: nil)
+        setNetworkResponseFor(url: FunctionalTestConst.EX_EDGE_INTERACT_PROD_URL_STR, httpMethod: HttpMethod.post, responseHttpConnection: responseConnection)
+        setExpectationNetworkRequest(url: FunctionalTestConst.EX_EDGE_INTERACT_PROD_URL_STR, httpMethod: HttpMethod.post, expectedCount: 1)
+
+        let experienceEvent = ExperienceEvent(xdm: ["testString": "xdm",
+                                                    "testInt": 10,
+                                                    "testBool": false,
+                                                    "testDouble": 12.89,
+                                                    "testArray": ["arrayElem1", 2, true],
+                                                    "testDictionary": ["key": "val"]])
```
emdobrin, line 976: let's create a constant for a default experienceEvent to save some lines

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
nporter-adbe: This isn't really possible unless we created our own initializer for the `EdgeEndpoint`. The default `rawValue` initializer that is provided for an enum with an associated type doesn't allow for `nil` argument and is also marked as a failable initializer.

I'd prefer not to overload the `rawValue` initializer to modify the behavior as it would become inconsistent with other enums in Swift and would be a net increase in overall lines of code.
https://developer.apple.com/documentation/swift/rawrepresentable/1538354-init

Tests/FunctionalTests/AEPEdgeFunctionalTests.swift
nporter-adbe: I would like to disable that rule, but I'd prefer to do it in a dedicated task/PR so that we can go through all the newly disabled rules and remove all the comment annotations throughout all our tests. Otherwise, I think we'll end up with many old unneeded `swiftlint:disable` comments.

I can create a task if you'd like.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
emdobrin: We do that for consentStatus, shouldn't be that bad, right? 
The main advantage of having that contained in one place is that in case the code is updated in the EdgeHitProcessor, the logic for determining the correct endpoint doesn't need to be impacted.
The processor should just extract the value from the config ss, but the enum will take care of returning the correct endpoint for that value.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -172,6 +174,9 @@ class EdgeHitProcessor: HitProcessing {
             return nil
         }
 
+        let edgeEndpointStr = configSharedState[EdgeConstants.SharedState.Configuration.EDGE_ENVIRONMENT] as? String
+        let edgeEndpoint = EdgeEndpoint(optionalRawValue: edgeEndpointStr?.lowercased())
```
emdobrin, line 178: can we move the lowercased inside EdgeEndpoint.init too?

emdobrin (approved):
Left a small comment, looks good otherwise

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
nporter-adbe: I don't mind lowercasing what we read from configuration, but I don't believe the actual enum should be case insensitive.

nporter-adbe: The main reason being is I don't think an enum case should match many different values, rather only the explicit string it defines.

The following behavior I'd also find undesired:
```swift
let rawValue = "INT"
let endpoint = EdgeEndpoint(optionalRawValue: rawValue)

let equal = endpoint.rawValue == rawValue // false
```

PR #219, opened 2021-12-01
Attach ImplementationDetails to XDM payload for each request

kevinlind:
## Description
Attaches ImplementationDetails to the XDM payload of each Experience Event.

See: [XDM ImplementationDetails schema](https://github.com/adobe/xdm/blob/014824a8b12d5828a6967781471f3a9765fcfbd1/components/datatypes/industry-verticals/implementationdetails.schema.json)
internal ticket: MOB-15365

## Types of changes

## Checklist:

### DIFF
--- a/Sources/Edge.swift
+++ b/Sources/Edge.swift
@@ -19,6 +19,7 @@ public class Edge: NSObject, Extension {
     private let SELF_TAG = "Edge"
     private var networkService: EdgeNetworkService = EdgeNetworkService()
     private var networkResponseHandler: NetworkResponseHandler?
+    private var implementationDetails: [String: Any]?
     internal var state: EdgeState?
 
     // MARK: - Extension
@@ -52,6 +53,9 @@ public class Edge: NSObject, Extension {
         registerListener(type: EventType.genericIdentity,
                          source: EventSource.requestReset,
                          listener: handleIdentitiesReset)
+        registerListener(type: EventType.hub,
+                         source: EventSource.sharedState,
+                         listener: handleSharedStateUpdate)
     }
 
     public func onUnregistered() {
@@ -175,6 +179,21 @@ public class Edge: NSObject, Extension {
         state?.hitQueue.queue(entity: entity)
     }
 
+    /// Handles shared state update events.
+    /// Updates `ImplementationDetails` if shared state event is from `EventHub`.
+    /// - Parameter event: current shared state update event
+    func handleSharedStateUpdate(_ event: Event) {
+        guard let data = event.data, !data.isEmpty else {
+            Log.trace(label: EdgeConstants.LOG_TAG, "\(SELF_TAG) - Shared State update event \(event.id.uuidString) cannot be processed as it contains no data.")
+            return
+        }
+
+        if data[EdgeConstants.SharedState.STATE_OWNER] as? String == EdgeConstants.SharedState.Hub.SHARED_OWNER_NAME {
+            let hubState = getSharedState(extensionName: EdgeConstants.SharedState.Hub.SHARED_OWNER_NAME, event: event)
+            implementationDetails = ImplementationDetails.from(hubState?.value ?? nil)
+        }
+    }
+
     /// Determines if the event should be ignored by the Edge extension. This method should be called after
     /// `readyForEvent` passed.
     ///
@@ -215,7 +234,8 @@ public class Edge: NSObject, Extension {
                                             networkResponseHandler: networkResponseHandler,
                                             getSharedState: getSharedState(extensionName:event:),
                                             getXDMSharedState: getXDMSharedState(extensionName:event:barrier:),
-                                            readyForEvent: readyForEvent(_:))
+                                            readyForEvent: readyForEvent(_:),
+                                            getImplementationDetails: getImplementationDetails)
         return PersistentHitQueue(dataQueue: dataQueue, processor: hitProcessor)
     }
 
@@ -229,6 +249,12 @@ public class Edge: NSObject, Extension {
         }
 
         return ConsentStatus.getCollectConsentOrDefault(eventData: consentXDMSharedState)
+    }
+
+    /// Returns current `ImplementationDetails` for this session.
+    /// - Returns: the `ImplementationDetails` for this session or nil if not yet set
+    private func getImplementationDetails() -> [String: Any]? {
+        return implementationDetails
     }
 
 }
--- a/Sources/Edge.swift
+++ b/Sources/Edge.swift
@@ -19,7 +19,7 @@ public class Edge: NSObject, Extension {
     private let SELF_TAG = "Edge"
     private var networkService: EdgeNetworkService = EdgeNetworkService()
     private var networkResponseHandler: NetworkResponseHandler?
-    private var implementationDetails: [String: Any]?
+    private var implementationDetails: [String: Any]? = ImplementationDetails.from(nil)
     internal var state: EdgeState?
 
     // MARK: - Extension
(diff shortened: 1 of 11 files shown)


Sources/Edge.swift
```diff
@@ -175,6 +179,21 @@ public class Edge: NSObject, Extension {
         state?.hitQueue.queue(entity: entity)
     }
 
+    /// Handles shared state update events.
+    /// Updates `ImplementationDetails` if shared state event is from `EventHub`.
+    /// - Parameter event: current shared state update event
+    func handleSharedStateUpdate(_ event: Event) {
```
emdobrin, line 185: Are we expecting the shared state update to always come first or should we attempt to read on bootup too?

emdobrin:
@kevinlind  can you please update the spec with an edge case section and provide an example for unknown?

Sources/ImplementationDetails.swift
```diff
@@ -0,0 +1,56 @@
...
+// the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+// OF ANY KIND, either express or implied. See the License for the specific language
+// governing permissions and limitations under the License.
+//
+
+import AEPCore
+import AEPServices
+import Foundation
+
+enum ImplementationDetails {
```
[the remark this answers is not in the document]
kevinlind: Using struct products the following warning from SwiftLint:

> Convenience Type Violation: Types used for hosting only static members should be implemented as a caseless enum to avoid instantiation. (convenience_type)

Sources/Edge.swift
kevinlind: Hmm.. I think the bigger issue is if Implementation Details is set before starting the hit queue, and here it is not. EdgeState.bootupIfNeeded is always called first as it is called from `readyForEvent`. I'll need to set the Implementation Details there, before EdgeState updates the hit queue with the initial Consent.

Sources/ImplementationDetails.swift
```diff
@@ -0,0 +1,56 @@
...
+
+        var wrapperName: String = ""
+
+        if hubState[EdgeConstants.SharedState.Hub.WRAPPER] != nil {
+            if let wrapperType = (hubState[EdgeConstants.SharedState.Hub.WRAPPER] as? [String: Any])?[EdgeConstants.SharedState.Hub.TYPE] as? String {
+                switch wrapperType {
+                case "R": // React Native
+                    wrapperName = "/\(EdgeConstants.JsonValues.ImplementationDetails.WRAPPER_REACT_NATIVE)" // note forward slash
+                default:
+                    wrapperName = "" // unsupported wrapper type defaults to none
```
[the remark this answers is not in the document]
kevinlind: Take a look at my recent changes after our discussion with Steve. I'm using the WrapperType enum to parse the type string from the shared state. WrapperType will default to None for unknown types.

Sources/ImplementationDetails.swift
```diff
@@ -0,0 +1,80 @@
...
+    /// - Parameter hubState: `EventHub` shared state
+    /// - Returns: wrapper type name used in the Implementation Details namespace, or an empty string if no wrapper is set
+    private static func getWrapperName(from hubState: [String: Any]) -> String {
+        if hubState[EdgeConstants.SharedState.Hub.WRAPPER] != nil {
+            let typeString = (hubState[EdgeConstants.SharedState.Hub.WRAPPER] as? [String: Any])?[EdgeConstants.SharedState.Hub.TYPE] as? String
+            if let wrapperTypeString = typeString, let wrapperType = WrapperType.init(rawValue: wrapperTypeString) {
+                return wrapperType == .none ? "" : "/\(wrapperType.stringValue)"
+            } else {
+                // "wrapper" entry exists but "type" not found. Unexpected, set to "unknown"
+                return "/\(EdgeConstants.JsonValues.ImplementationDetails.UNKNOWN)" // note forward slash
```
emdobrin, line 53: Let's update this method to only return the wrapper type string L50, L53 and append the slash when used

Tests/UnitTests/EdgeHitProcessorTests.swift
```diff
@@ -456,6 +464,108 @@ class EdgeHitProcessorTests: XCTestCase {
...
+
+        let payload = asFlattenDictionary(data: requestPayload)
+
+        XCTAssertEqual("https://ns.adobe.com/experience/mobilesdk/ios", payload["xdm.implementationDetails.name"] as? String)
+        XCTAssertEqual("3.3.1+1.0.0", payload["xdm.implementationDetails.version"] as? String)
+        XCTAssertEqual("app", payload["xdm.implementationDetails.environment"] as? String)
+    }
+
+    // tests Implementation Details is not added to event when nil
+    func testProcessHit_experienceEvent_sendsNetworkRequest_withOutImplementationDetails_whenNil() {
```
emdobrin, line 505: typo: without

Tests/UnitTests/EdgeHitProcessorTests.swift
```diff
@@ -456,6 +464,108 @@ class EdgeHitProcessorTests: XCTestCase {
...
+
+        let payload = asFlattenDictionary(data: requestPayload)
+
+        XCTAssertNil(payload["xdm.implementationDetails.name"])
+        XCTAssertNil(payload["xdm.implementationDetails.version"])
+        XCTAssertNil(payload["xdm.implementationDetails.environment"])
+    }
+
+    // tests Implementation Details is not added to Consent events
+    func testProcessHit_consentEvent_sendsNetworkRequest_withoutImplementationDetails() {
```
emdobrin, line 534: Doc: would you include this in the spec

Tests/UnitTests/ImplementationDetailsTests.swift
```diff
@@ -0,0 +1,290 @@
...
+
+        let expectedResult: [String: Any] = [
+            "name": "\(BASE_NAMESPACE)/\(WRAPPER_REACT_NATIVE)",
+            "environment": "app",
+            "version": "3.3.1+\(EdgeConstants.EXTENSION_VERSION)"
+        ]
+        assertEqual(expectedResult, actualResult)
+    }
+
+    func testFrom_withVersion_withWrapperFlutter() {
```
emdobrin, line 72: would be nice to have a test configuration for these repeating tests and switch the wrapperType val and expected namespace, would reduce this file lines significantly.

Sources/ImplementationDetails.swift
kevinlind: I'm not clear on what you are asking. Currently, the method will return:

- "" (empty string) if wrapper type is None or wrapper is not found in shared state
- "/unknown" if wrapper type is not found in shared state
- "/<wrapper type>"

emdobrin: E.g. instead of /wrappertypevalue can we just return wrappertypevalue and let the caller to attach the forward slash

Sources/Edge.swift
```diff
@@ -231,4 +232,10 @@ public class Edge: NSObject, Extension {
         return ConsentStatus.getCollectConsentOrDefault(eventData: consentXDMSharedState)
     }
 
+    /// Returns current `ImplementationDetails` for this session.
+    /// - Returns: the `ImplementationDetails` for this session or nil if not yet set
+    private func getImplementationDetails() -> [String: Any]? {
+        return state?.implementationDetails
```
emdobrin, line 238: since you are calling this from the hit processor, you might want to protect the access to it

emdobrin (commented):
looks good with a threading comment

emdobrin (approved):

PR #235, opened 2022-03-21
[MOB-15894] Format Experience Event timestamps in UTC with millisecond precision

kevinlind:
## Description

Use AEPServices Date+Format to format Experience Event timestamps as ISO 8601 dates in UTC with millisecond precision.
Add deprecated warning to XDMFormatters functions with message to use Date+Format instead.

IMPORTANT: Changed AEPCore location in Podfile to use Git branch `dev-v3.4.3`. The Podfile will need to be updated once AEPCore production is released but before AEPEdge is released.

Screen Shot 2022-03-21 at 1 18 51 PM

## Types of changes

## Checklist:

### DIFF
--- a/Sources/xdm/XDMFormatters.swift
+++ b/Sources/xdm/XDMFormatters.swift
@@ -22,6 +22,7 @@ public enum XDMFormatters {
     ///   - Date: A timestamp and it must not be null
     /// - Returns: The timestamp formatted to a string in the format of 'yyyy-MM-dd'T'HH:mm:ssXXX',
     ///            or an empty string if Date  is null
+    @available(*, deprecated, message: "Use function getISO8601UTCDateWithMilliseconds() in Date class extension from AEPServices module instead.")
     public static func dateToISO8601String(from: Date?) -> String? {
         if let unwrapped = from {
             return unwrapped.asISO8601String()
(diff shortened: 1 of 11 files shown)


Sources/xdm/XDMFormatters.swift
```diff
@@ -22,6 +22,7 @@ public enum XDMFormatters {
     ///   - Date: A timestamp and it must not be null
     /// - Returns: The timestamp formatted to a string in the format of 'yyyy-MM-dd'T'HH:mm:ssXXX',
     ///            or an empty string if Date  is null
+    @available(*, deprecated, message: "Use function getISO8601UTCDateWithMilliseconds() in Date class extension from AEPServices module instead.")
```
emdobrin, line 25: This seems ok to me. I wish we could have used the renamed attribute here but the usage is slightly different and in a different module.

emdobrin (commented):
Can you please also update the XDMFormatters usages in the test apps, test schemas and any other occurrences?

Sources/xdm/XDMFormatters.swift
kevinlind: I played around with `renamed` but it appears the method needs to be in the same class as that of the deprecated. Plus, as you said, the replacement is an extension to Date and not a utility function, so the usage is different.
There's also the `obsoleted` tag, but it appears that applies to either a Swift or iOS version, and not the SDK version, so that doesn't make much sense to use in this case.

emdobrin (approved):

PR #243, opened 2022-04-14
Add query operation update to consent requests

emdobrin:
## Description

MOB-15907 - Adds query operation update setting to all consent requests in preparation for incremental updates for adid consent.

## Motivation and Context

The query option is required to overwrite the default behavior in Edge Network where collect consent is required with the consent preferences. This allows for other consent preferences to be collected earlier in the application lifecycle, while the default collect settings may be used.

## Types of changes

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.


kevinlind (approved):
Nice! I'm not finding any faults with this.

kevinlind:
> @kevinlind thanks for reviewing. The build was failing and I made a small change - reverted the pod updates. It looks like the latest AEPEdgeConsent brings AEPEdge as a dependency and once installed it shows as a new scheme in Xcode so the xcodebuild gets confused to which one to pick. We have couple options for this:
> 
> * We clean up the test apps and tests in this repo to not bring external dependencies other than AEPCore so we can continue using AEPEdge from source code. Any integration tests with Consent + Edge will be moved to Consent extension. Any manual integration tests will be executed using the sample app.
> * We define a test podspec for AEPEdgeConsent and exclude the AEPEdge dependency so we can continue using it.
> 
> I am inclined towards the first one, let me know your thoughts.

I'm inclined to the first option as well. However, I'd like to know more about change https://github.com/adobe/aepsdk-edgeconsent-ios/pull/43 to see if we can simply remove the AEPEdge dependency from the AEPEdgeConsent podspec.

emdobrin:
Even though there is no hard dependency on the Edge extension APIs, the Consent extension relies on the Edge functionality for handling the consent updates. Any new feature or improvement to support the Consent workflows will require a bump in this dependency, and I think the dependency link is still required for better version management.
I created a new task for the clean-up.

PR #245, opened 2022-04-26
[DEV] Updates on EdgeEndpoint, renaming and clean up on unused fields

emdobrin:
## Description

- Updates how EdgeEndpoint is calculated using URLComponents and requires EdgeRequestType
- Renames ExperienceEdgeRequestType -> EdgeRequestType
- Removes unused EdgeRequestType.collect, can be re-added in the future if needed
- Group test cases by area to reduce lines of code in EdgeEndpoint tests, adds test for consent endpoints

## Motivation and Context

Changes made in preparation for path overwrite

## Types of changes

## Checklist:

### DIFF
--- a/Sources/EdgeNetworkHandlers/EdgeEndpoint.swift
+++ b/Sources/EdgeNetworkHandlers/EdgeEndpoint.swift
@@ -33,13 +33,16 @@ enum EdgeEnvironmentType: String {
 }
 
 struct EdgeEndpoint {
-    let endpointUrl: String
+    let url: URL?
 
     /// Initializes the appropriate `EdgeEndpoint` for the given `type` and `optionalDomain`
     /// - Parameters:
-    ///   - type: the `EdgeEnvironmentType` for the `EdgeEndpoint`
+    ///   - requestType: the `EdgeRequestType` to be used
+    ///   - environmentType: the `EdgeEnvironmentType` for the `EdgeEndpoint`
     ///   - optionalDomain: an optional custom domain for the `EdgeEndpoint`. If not set the default domain is used.
-    init(type: EdgeEnvironmentType, optionalDomain: String? = nil) {
+    init(requestType: EdgeRequestType,
+         environmentType: EdgeEnvironmentType,
+         optionalDomain: String? = nil) {
         let domain: String
         if let unwrappedDomain = optionalDomain, !unwrappedDomain.isEmpty {
             domain = unwrappedDomain
@@ -47,14 +50,25 @@ struct EdgeEndpoint {
             domain = EdgeConstants.NetworkKeys.EDGE_DEFAULT_DOMAIN
         }
 
-        switch type {
+        var components = URLComponents()
+        components.scheme = EdgeConstants.NetworkKeys.HTTPS
+
+        switch environmentType {
         case .production:
-            endpointUrl = "https://\(domain)\(EdgeConstants.NetworkKeys.EDGE_ENDPOINT_PATH)"
+            components.host = domain
+            components.path = EdgeConstants.NetworkKeys.EDGE_ENDPOINT_PATH
         case .preProduction:
-            endpointUrl = "https://\(domain)\(EdgeConstants.NetworkKeys.EDGE_ENDPOINT_PRE_PRODUCTION_PATH)"
+            components.host = domain
+            components.path = EdgeConstants.NetworkKeys.EDGE_ENDPOINT_PRE_PRODUCTION_PATH
         case .integration:
             // Edge Integration endpoint does not support custom domains, so there is just the one URL
-            endpointUrl = EdgeConstants.NetworkKeys.EDGE_ENDPOINT_INTEGRATION
+            components.host = EdgeConstants.NetworkKeys.EDGE_INTEGRATION_DOMAIN
+            components.path = EdgeConstants.NetworkKeys.EDGE_ENDPOINT_PATH
         }
+
+        components.path.append(EdgeConstants.NetworkKeys.EDGE_ENDPOINT_VERSION_PATH)
+        components.path.append("/\(requestType.rawValue)")
+
+        url = components.url
     }
 }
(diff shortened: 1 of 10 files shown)


Sources/EdgeNetworkHandlers/EdgeEndpoint.swift
```diff
@@ -33,28 +33,42 @@ enum EdgeEnvironmentType: String {
...
+            components.path = EdgeConstants.NetworkKeys.EDGE_ENDPOINT_PRE_PRODUCTION_PATH
         case .integration:
             // Edge Integration endpoint does not support custom domains, so there is just the one URL
-            endpointUrl = EdgeConstants.NetworkKeys.EDGE_ENDPOINT_INTEGRATION
+            components.host = EdgeConstants.NetworkKeys.EDGE_INTEGRATION_DOMAIN
+            components.path = EdgeConstants.NetworkKeys.EDGE_ENDPOINT_PATH
         }
+
+        components.path.append(EdgeConstants.NetworkKeys.EDGE_ENDPOINT_VERSION_PATH)
+        components.path.append("/\(requestType.rawValue)")
```
addb, line 70: should we have a constant for this? 

`let EDGE_REQUEST_TYPE = "/\(requestType.rawValue)"`

so we can use that constant to add to components like we do for all the other components
`components.path.append(EDGE_REQUEST_TYPE)`

Sources/EdgeNetworkHandlers/EdgeHit.swift
```diff
@@ -25,8 +25,8 @@ protocol EdgeHit {
     /// Unique identifier for this hit
     var requestId: String { get }
```
addb, line 26: nit: How about renaming the `EdgeHit` to `EdgeRequest`
also renaming several fields
`edgeEndpoint` to `endpoint`
`requestId` to `id`

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -129,6 +131,17 @@ class EdgeHitProcessor: HitProcessing {
         }
     }
 
+    /// Builds the endpoint based on the provided config info and `EdgeRequestType`
+    /// - Parameters:
+    ///   - config: configuration data, used to extract the environment and the custom domain, if any
+    ///   - requestType: the `EdgeRequestType`
+    private func buildEndpoint(config: [String: String], requestType: EdgeRequestType) -> EdgeEndpoint {
```
addb, line 138: nit: buildEdgeEndpoint and we can drop edge from the function parameters.

Sources/EdgeNetworkHandlers/EdgeHit.swift
addb: I noticed we have EdgeRequests as well, then should we rename requestId, to just Id Which will translate to hitId. If we have request and hit as two different terms, I feel we can do some cleanup to avoid any confusions.

Tests/UnitTests/EdgeEndpointTests.swift
```diff
@@ -56,49 +57,59 @@ class EdgeEndpointTests: XCTestCase {
...
 
-    func testDefaultProductionEndpoint() {
-        let endpoint = EdgeEndpoint(type: .production)
-        let expected = "https://\(EdgeConstants.NetworkKeys.EDGE_DEFAULT_DOMAIN)\(EdgeConstants.NetworkKeys.EDGE_ENDPOINT_PATH)"
-        XCTAssertEqual(expected, endpoint.endpointUrl)
-    }
+    // MARK: EdgeEndpoint tests
+    func testEdgeEndpoint_interact_defaultDomain() {
+        // cases defined as: ($0: input EnvironmentType, $1: expected output EdgeEndpoint URL, $2: Test case name)
+        let cases = [(EdgeEnvironmentType.production, "https://edge.adobedc.net/ee/v1/interact", "DefaultProductionEndpoint"),
```
kevinlind, line 63: This will work better if the test case doesn't fail after the first failure. Looking at the other tests in this test suite, it looks like you can set `continueAfterFailure = true`, but you should double check. I'd suggest explicitly setting the flag to true with a comment stating it allows these tests with multiple parameters to run each case even when one fails, so someone later doesn't set it to false.

Sources/EdgeNetworkHandlers/EdgeHit.swift
```diff
@@ -25,8 +25,8 @@ protocol EdgeHit {
...
 
-    /// The `ExperienceEdgeRequestType` to be used for this `EdgeHit`
-    func getType() -> ExperienceEdgeRequestType
+    /// The `EdgeRequestType` to be used for this `EdgeHit`
+    func getType() -> EdgeRequestType
```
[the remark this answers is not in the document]
emdobrin: Good catch

Sources/EdgeNetworkHandlers/EdgeHit.swift
emdobrin: I updated the comment for `requestId` and renamed field edgeEndpoint -> endpoint as the class name contains Edge already, usage example:
```
ExperienceEventsEdgeHit(endpoint: EdgeEndpoint(...), configId: configId, request: requestPayload)
```

Sources/EdgeNetworkHandlers/EdgeEndpoint.swift
emdobrin: Personally I would prefer not to create a variable just for this unless you feel this line is not clear for other reasons.

kevinlind (approved):

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
emdobrin: params seem alright, is your comment to rename buildEndpoint -> buildEdgeEndpoint?

addb: yes

addb (approved):

PR #258, opened 2022-08-02
[MOB-16479] Allow path overwrite for Media Edge requests

addb:
## Description
Add support to override interact request path. This is primarily being added to support Media edge workflow, but is generic, so that any other extension can make use of it. 

Format | https:{EdgeDomain}/{Environment}/{EdgeRegionId}/{custom path}
-- | --
Example | https://edge.domain/ee/or2/va/v1/sessionStart

## Types of changes

## Checklist:


Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -224,4 +226,15 @@ class EdgeHitProcessor: HitProcessing {
         return requestHeaders
     }
 
+    private func getCustomPathFromEvent(_ event: Event) -> String? {
```
kevinlind, line 229: nit: Rename to `getCustomRequestPath(from event: Event) -> String?`. Then call by `getCustomRequestPath(from: myEvent)`

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -97,7 +97,8 @@ class EdgeHitProcessor: HitProcessing {
                 return
             }
 
-            let endpoint = buildEdgeEndpoint(config: edgeConfig, requestType: EdgeRequestType.interact)
+            let overwritePath = getCustomPathFromEvent(event)
+            let endpoint = buildEdgeEndpoint(config: edgeConfig, requestType: EdgeRequestType.interact, overwritePath: overwritePath)
```
kevinlind, line 101: To clean this up, I'd suggest calling `getCustomPathFromEvent` in `getEdgeConfig` and store the custom request path in the `edgeConfig` dictionary. You won't need to change the `buildEdgeEndpoint` signature that way as the `config` parameter will contain the custom path.

Sources/EdgeNetworkHandlers/EdgeEndpoint.swift
```diff
@@ -42,7 +42,8 @@ struct EdgeEndpoint {
     ///   - optionalDomain: an optional custom domain for the `EdgeEndpoint`. If not set the default domain is used.
     init(requestType: EdgeRequestType,
          environmentType: EdgeEnvironmentType,
-         optionalDomain: String? = nil) {
+         optionalDomain: String? = nil,
+         overwritePath: String? = nil) {
```
kevinlind, line 46: To keep in line with the naming of the other parameters, maybe rename to `optionalPath` ?

Sources/EdgeNetworkHandlers/EdgeEndpoint.swift
```diff
@@ -42,7 +42,8 @@ struct EdgeEndpoint {
     ///   - optionalDomain: an optional custom domain for the `EdgeEndpoint`. If not set the default domain is used.
     init(requestType: EdgeRequestType,
```
kevinlind, line 43: Add documentation for new parameter.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
kevinlind: I still see it as a configuration value for the request but where the configuration is coming from the event data instead of the shared state. You can also update the function description to something more generic like "Extracts all the Edge configuration keys needed for this request. Configuration keys are extracted from the Configuration shared state and the given 'event' instance."

Sources/EdgeNetworkHandlers/EdgeEndpoint.swift
```diff
@@ -40,9 +40,11 @@ struct EdgeEndpoint {
     ///   - requestType: the `EdgeRequestType` to be used
     ///   - environmentType: the `EdgeEnvironmentType` for the `EdgeEndpoint`
     ///   - optionalDomain: an optional custom domain for the `EdgeEndpoint`. If not set the default domain is used.
+    ///   - optionalPath: an optional path to be used to overwrite the default interact path.
```
kevinlind, line 43: As the endpoint can be for interact, collect, or consent requests, can you reword to "... overwrite the default path".

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -181,7 +182,7 @@ class EdgeHitProcessor: HitProcessing {
         }
     }
 
-    /// Extracts all the Edge configuration keys from the Configuration shared state
+    /// Extracts all the Edge configuration keys needed for this request. Configuration keys are extracted from the `Configuration shared state` and the given `event` instance
```
kevinlind, line 185: You don't need the back ticks around "Configuration shared state" here as it adds extra highlighting to the text. The reason why I suggested adding them to "event" is because it's the name of the parameter. For example, in Javadoc, it's common to refer to a parameter name with `{@code event}`, where Swift uses back ticks instead of the code annotation. 
Also add a period at the end of the sentence.

Tests/UnitTests/EdgeHitProcessorTests.swift
```diff
@@ -552,6 +583,10 @@ class EdgeHitProcessorTests: XCTestCase {
         XCTAssertNil(payload["xdm.implementationDetails.environment"])
     }
 
+    func assertHitPath() {
+        mockNetworkService
+    }
```
kevinlind, line 588: What does this function do? I don't see that it is called in this file.

kevinlind:
Do you know why the coverage drops 2.5% with these changes?  The report states that your changes are 100% covered by tests, however coverage is lost on lines this PR doesn't change.

addb:
> Do you know why the coverage drops 2.5% with these changes?  The report states that your changes are 100% covered by tests, however coverage is lost on lines this PR doesn't change. 

It has to do with the Xcode 13.0.0 update. We have noticed this in other repos as well that the coverage drops when we update to xcode 13.0.0

Sources/EdgeConstants.swift
```diff
@@ -99,6 +103,8 @@ enum EdgeConstants {
         static let EVENT_ID = "_id"
         static let META = "meta"
         static let IMPLEMENTATION_DETAILS = "implementationDetails"
+        static let REQUEST = "request"
+        static let PATH = "path"
```
emdobrin, line 108: These JSONKeys are used mainly for the network request body, let's keep just the EventDataKeys ones.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -181,7 +182,7 @@ class EdgeHitProcessor: HitProcessing {
         }
     }
 
-    /// Extracts all the Edge configuration keys from the Configuration shared state
+    /// Extracts all the Edge configuration keys needed for this request. Configuration keys are extracted from the Configuration shared state and the given `event` instance.
```
emdobrin, line 185: I recommend not mixing config keys with event keys in the same payload here to keep this function code specialized on a single item at a time. We could instead use a new helper `getRequestProperties(event)` that extract the request path overwrite and maybe additional request settings in the future.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -224,4 +229,15 @@ class EdgeHitProcessor: HitProcessing {
         return requestHeaders
     }
 
+    private func getCustomRequestPath(from event: Event) -> String? {
+        var customPath: String?
+        if let eventData = event.data {
+            let xdmData = eventData[EdgeConstants.JsonKeys.XDM] as? [String: Any]
```
emdobrin, line 235: the new "request" key should be in the event data outside of the xdm payload as per spec

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
emdobrin: Additionally getEdgeConfig is used for all the endpoints we integrate with (currently interact and consent), we should allow the path overwrite only in the edge requestContent (interact) requests, and not for consent. Let's also include a test that overwrites are ignored for consent

emdobrin (commented):
@addb can you also update the description for this PR?

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -224,4 +229,15 @@ class EdgeHitProcessor: HitProcessing {
...
+            let requestData = xdmData?[EdgeConstants.JsonKeys.REQUEST] as? [String: Any]
+            customPath = requestData?[EdgeConstants.JsonKeys.PATH] as? String
```
emdobrin, line 237: Usually a path starts with slash, so I am thinking we should require that like `/va/v1/sessionStart`. We could also verify that it does not start multiple slashes (// is used before the domain/authority component) and if so ignore and log a debug message, though can you verify how will path.path.append behave in that case e.g. //va/v1?

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
addb: Oh yes.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -229,15 +226,46 @@ class EdgeHitProcessor: HitProcessing {
...
 
         return customPath
     }
 
+    private func isValidPath(_ path: String?) -> Bool {
+        guard let path = path else {
+            return false
+        }
+
+        let pattern = "^\\/[/.a-zA-Z0-9-]+$"
```
kevinlind, line 261: Let's allow all the unreserved characters in a URL, so add underscore `_` and tilde `~`.
Unreserved characters in URL include upper and lower Alpha, Numeric, plus `.`, `_`, `-`, and `~`.
See https://www.ietf.org/rfc/rfc3986.txt section 2.3

Tests/UnitTests/EdgeHitProcessorTests.swift
```diff
@@ -50,9 +50,23 @@ class EdgeHitProcessorTests: XCTestCase {
...
+    let experienceEventWithOverwritePath = Event(name: "test-experience-event", type: EventType.edge, source: EventSource.requestContent, data: ["xdm": ["test": "data"], "request": ["path": "/va/v1/sessionstart"]])
+
+    let invalidPaths = [
+        "/va/v1/sessionstart?query=value",
+        "//va/v1/sessionstart",
+        "/va/v1//sessionstart",
+        "/va/v1/sessionstart/@test"
+    ]
+
+    let validPaths = [
```
kevinlind, line 62: Add test paths which include the underscore and tilde characters after adding them to the allowed character pattern in my previous comment.

kevinlind (commented):
Can you check the coverage report and increase your diff coverage back to 100%? You may just need to add some negative tests to `isValidPath`.

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -241,23 +241,23 @@ class EdgeHitProcessor: HitProcessing {
...
 
     private func isValidPath(_ path: String?) -> Bool {
         guard let path = path else {
             return false
         }
 
+        if path.isEmpty || path.contains("//") {
+            return false
+        }
+
```
kevinlind, line 260: Could you combine this with the above, like?
```
guard let path = path, !path.isEmpty, !path.contains("//") else {
    return false
}
```

kevinlind (approved):
One additional small comment, otherwise looks good.

Sources/EdgeNetworkHandlers/EdgeEndpoint.swift
```diff
@@ -64,10 +66,16 @@ struct EdgeEndpoint {
             // Edge Integration endpoint does not support custom domains, so there is just the one URL
             components.host = EdgeConstants.NetworkKeys.EDGE_INTEGRATION_DOMAIN
             components.path = EdgeConstants.NetworkKeys.EDGE_ENDPOINT_PATH
+
         }
 
-        components.path.append(EdgeConstants.NetworkKeys.EDGE_ENDPOINT_VERSION_PATH)
-        components.path.append("/\(requestType.rawValue)")
+        if let customPath = optionalPath {
+            // path should contain the leading "\"
```
emdobrin, line 73: this should probably be `the leading slash "/"`

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
emdobrin: Should we revert this doc change?

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -224,4 +226,39 @@ class EdgeHitProcessor: HitProcessing {
         return requestHeaders
     }
 
+    private func getRequestProperties(from event: Event) -> [String: Any]? {
```
emdobrin, line 229: missing docs on the functions below

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -224,4 +226,39 @@ class EdgeHitProcessor: HitProcessing {
...
+
+    private func getCustomRequestPath(from event: Event) -> String? {
+        var path: String?
+        if let eventData = event.data {
+            let requestData = eventData[EdgeConstants.EventDataKeys.Request.KEY] as? [String: Any]
+            path = requestData?[EdgeConstants.EventDataKeys.Request.PATH] as? String
+        }
+
+        if !isValidPath(path) {
+            Log.error(label: self.SELF_TAG, "Dropping the overwrite path value: (\(String(describing: path))), since it contains invalid characters or is empty.")
```
emdobrin, line 245: String (describing..) creates a new String. we can use '\\(path ?? "")'

Tests/FunctionalTests/EdgePathOverwriteTests.swift
```diff
@@ -0,0 +1,156 @@
...
+
+        let experienceEventWithOverwritePath = Event(name: "test-experience-event", type: EventType.edge, source: EventSource.requestContent, data: ["xdm": ["test": "data"], "request": ["path": "/va/v1/sessionstart"]])
+        MobileCore.dispatch(event: experienceEventWithOverwritePath)
+
+        // verify
+        assertNetworkRequestsCount()
+        let resultNetworkRequests = getNetworkRequestsWith(url: FunctionalTestConst.EX_EDGE_MEDIA_PROD_URL_STR, httpMethod: HttpMethod.post)
+
+        let requestUrl = resultNetworkRequests[0].url
+        XCTAssertTrue(requestUrl.absoluteURL.absoluteString.hasPrefix(FunctionalTestConst.EX_EDGE_MEDIA_PROD_URL_STR))
```
emdobrin, line 77: Per spec you could use `requestUrl.absoluteString` > `The URL string for the receiver as an absolute URL` https://developer.apple.com/documentation/foundation/nsurl/1409868-absolutestring
Or you could just extract the path portion with requestUrl.path
Applies for all usages below

Tests/UnitTests/EdgeHitProcessorTests.swift
```diff
@@ -278,6 +302,51 @@ class EdgeHitProcessorTests: XCTestCase {
...
+    }
+
+    func testProcessHit_withOverwritePath_invalidPath_doesNotOverwriteThePath() {
+        for path in invalidPaths {
+            let expEventInvalidPath = Event(name: "test-consent-event", type: EventType.edge, source: EventSource.requestContent, data: ["request": ["path": path]])
+            assertNetworkRequestUrl(event: expEventInvalidPath, environment: "prod", domain: nil, expectedEndpoint: INTERACT_ENDPOINT_PROD)
+        }
+    }
+
+    func testProcessHit_consentUpdateEvent_withOverwritePath_doesNotOverwriteThePath() {
```
emdobrin, line 346: nit: `testProcessHit_consentUpdateEvent_withOverwritePath_ignoresOverwrite`

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -224,4 +226,39 @@ class EdgeHitProcessor: HitProcessing {
...
+
+        return path
+    }
+
+    private func isValidPath(_ path: String?) -> Bool {
+        guard let path = path, !path.isEmpty, !path.contains("//") else {
+            return false
+        }
+
+        let pattern = "^\\/[/.a-zA-Z0-9-~_]+$"
```
emdobrin, line 257: extract as constant

Tests/FunctionalTests/EdgePathOverwriteTests.swift
```diff
@@ -0,0 +1,156 @@
...
+
+@testable import AEPCore
+@testable import AEPEdge
+import AEPEdgeIdentity
+import AEPServices
+import Foundation
+import XCTest
+
+/// End-to-end testing for the AEPEdge public APIs
+class AEPEdgePathOverwriteTests: FunctionalTestBase {
```
emdobrin, line 21: can you also add a test that it does not include the "request" portion in the request body?

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -242,7 +248,7 @@ class EdgeHitProcessor: HitProcessing {
         }
 
         if !isValidPath(path) {
-            Log.error(label: self.SELF_TAG, "Dropping the overwrite path value: (\(String(describing: path))), since it contains invalid characters or is empty.")
+            Log.error(label: self.SELF_TAG, "Dropping the overwrite path value: (\(path ?? "")), since it contains invalid characters or is empty.")
```
emdobrin, line 251: We might need to split isValidPath into two checks, we should only error out if there is an invalid path, but not for nil/empty as that is the default scenario.

Tests/FunctionalTests/EdgePathOverwriteTests.swift
addb: I see the request being sent as event payload.

emdobrin: hmm.. that should not be the case, we should handle this and not include the client-side "request" portion in the request to Konductor

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
kevinlind: Don't forget to revert this doc change.

kevinlind (approved):
Looks good. One small clean up comment before you merge to revert a doc change after you refactored some code.

PR #259, opened 2022-08-29
[MOB-16678] Handle edge network location hint response

kevinlind:
## Description

Handle the EdgeNetwork location hint response from Konductor.

Reads location hint and ttl from Konductor response and stores in EdgeProperties. All requests from the Edge extension will include the location hint in the request URL if available and if not expired. The location hint and expiry date (calculated from the ttlSeconds) is persisted in User Defaults and loaded once the Edge extension launches.  
The Edge extension now creates a shared state which includes just the location hint. A shared state is created on launch of the extension and when ever the location hint value changes. A shared state is not created, however, if only the expiry date is updated but the hint value doesn't change.

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ jobs:
 
   build-and-test:
     macos:
-      xcode: 12.0.1 # Specify the Xcode version to use
+      xcode: 13.4.0 # Specify the Xcode version to use
 
     steps:
       - checkout
--- a/Sources/Edge.swift
+++ b/Sources/Edge.swift
@@ -241,6 +241,7 @@ public class Edge: NSObject, Extension {
 
     /// Set the location hint for the Edge Network. The new location hint and expiry date (calculated from the ttlSeconds) are updated in memory and in
     /// persistent storage. If the new location hint is different from the previous, then a shared state is also created with the new hint.
+    /// A nil `hint` value will clear the location hint in memory, persistent storage, and create a new shared state if a previous location hint was set.
     /// - Parameters:
     ///   - hint: the new EdgeNetwork location hint to set
     ///   - ttlSeconds: the time-to-live for the location hint
@@ -248,6 +249,8 @@ public class Edge: NSObject, Extension {
         guard let state = state else { return }
         if let hint = hint, let ttlSeconds = ttlSeconds {
             state.setLocationHint(hint: hint, ttlSeconds: ttlSeconds, createSharedState: createSharedState(data:event:))
+        } else if hint == nil {
+            state.clearLocationHint(createSharedState: createSharedState(data:event:))
         }
     }
 
--- a/Sources/Edge.swift
+++ b/Sources/Edge.swift
@@ -35,7 +35,7 @@ public class Edge: NSObject, Extension {
         // set default on init for register/unregister use-case
         networkResponseHandler = NetworkResponseHandler(updateLocationHint: setLocationHint)
         if let hitQueue = setupHitQueue() {
-            state = EdgeState(hitQueue: hitQueue)
+            state = EdgeState(hitQueue: hitQueue, edgeProperties: EdgeProperties())
         }
     }
 
@@ -195,7 +195,9 @@ public class Edge: NSObject, Extension {
     /// - Returns: true if events can be processed at the moment, false otherwise
     private func canProcessEvents(event: Event) -> Bool {
         guard let state = state else { return false }
-        state.bootupIfNeeded(event: event, getSharedState: getSharedState(extensionName:event:barrier:))
+        state.bootupIfNeeded(event: event,
+                             getSharedState: getSharedState(extensionName:event:barrier:),
+                             createSharedState: createSharedState(data:event:))
         return true
     }
 
--- a/Sources/Edge.swift
+++ b/Sources/Edge.swift
@@ -33,7 +33,7 @@ public class Edge: NSObject, Extension {
         super.init()
 
         // set default on init for register/unregister use-case
-        networkResponseHandler = NetworkResponseHandler()
+        networkResponseHandler = NetworkResponseHandler(updateLocationHint: setLocationHint)
         if let hitQueue = setupHitQueue() {
             state = EdgeState(hitQueue: hitQueue)
         }
@@ -216,7 +216,8 @@ public class Edge: NSObject, Extension {
                                             getSharedState: getSharedState(extensionName:event:),
                                             getXDMSharedState: getXDMSharedState(extensionName:event:barrier:),
                                             readyForEvent: readyForEvent(_:),
-                                            getImplementationDetails: getImplementationDetails)
+                                            getImplementationDetails: getImplementationDetails,
+                                            getLocationHint: getLocationHint)
         return PersistentHitQueue(dataQueue: dataQueue, processor: hitProcessor)
     }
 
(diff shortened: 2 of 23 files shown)


Sources/EdgeState.swift
```diff
@@ -62,6 +75,9 @@ class EdgeState {
         }
         // else keep consent pending until the consent preferences update event is received
 
+        // TODO - Always set state or only when properties are not empty ???
+        createSharedState(edgeProperties.toEventData(), nil)
```
kevinlind, line 79: Always creates a shared state when EdgeState boots, regardless of location hint value. Another option is to only create the shared state if there is a valid location hint value, but I think creating an initial empty shared state on boot is better approach and more consistent with how other extensions are implemented.

Sources/EdgeState.swift
```diff
@@ -74,4 +90,57 @@ class EdgeState {
...
+
+            if needsStateUpdate {
+                // Create shared state if location hint changed
+                // Important - Using nil Event here which creates a shared state at the next available Event number.
+                //             An extension should NOT mix creating shared states using nil and using received events
+                //             as it can cause shared state generation to fail due to received events having potentially
+                //             lower event numbers than states using nil. If this extension later needs to create shared
+                //             states from received events, then this code must be refactored to also use received
+                //             events as the state version.
+                createSharedState(self.edgeProperties.toEventData(), nil)
```
kevinlind, line 121: I'm using `nil` here for the shared state version, which will cause the Event Hub to get the next available version number. I'm using nil because this function is called by the NetworkResponseHandler which does not have a triggering event to create a shared state with. Using nil here means I must always use nil when creating a shared state even when I have one available (such as in bootupIfNeeded). The shared state is created mainly for Rules use-cases where a customer may want to build a postback URL using the location hint.

Another option is for NetworkResponseHandler to dispatch an event which is listened for by Edge to handle the location hint. That way there is an event to use to create the shared state. This is how Identity Direct extension handles this case. This option, however, will cause a delay when the location hint is handled and any queued hits will not include the hint in their URLs.

 As there are not current plans for Edge to include other properties in it's shared state, I feel the first option of using 'nil' is the better approach, but I wanted to point this out to hear other's thoughts.

.circleci/config.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
   build-and-test:
     macos:
-      xcode: 12.0.1 # Specify the Xcode version to use
+      xcode: 13.4.0 # Specify the Xcode version to use
```
addb, line 17: Is the code coverage issue fixed in 13.4?

kevinlind: I'm not seeing an issue with CodeCov. I've updated the version because version 12 of Xcode is no longer supported in CircleCI. The builds just fail now. Their documentation states that they support the latest 4 major.minor versions of Xcode. We've used Xcode 13.0.0 for other projects, but that is the 4th version. If there is a 13.5 version of Xcode, then 13.0.0 may not be supported. I didn't want to bump all the way to Xcode 14, so I picked that latest Xcode 13 image supported by CircleCI.

Sources/EdgeConstants.swift
```diff
@@ -117,11 +122,18 @@ enum EdgeConstants {
...
             static let EVENT_HANDLE_TYPE_STORE = "state:store"
+            static let EVENT_HANDLE_TYPE_LOCATION_HINT = "locationHint:result"
 
             enum Error {
                 static let MESSAGE = "message"
                 static let NAMESPACE = "namespace"
             }
+
+            enum LocationHint {
+                static let SCOPE = "scope"
```
addb, line 133: Should we add the "locationHint" constant here only? 
static let KEY = "locationHint"

Sources/EdgeConstants.swift
```diff
@@ -117,11 +122,18 @@ enum EdgeConstants {
 
         enum Response {
             static let EVENT_HANDLE_TYPE_STORE = "state:store"
```
addb, line 124: Does it make sense to have enum for EventHandle?

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -317,6 +325,23 @@ class NetworkResponseHandler {
...
 
+    /// If handle is of type "locationHint:result", persist it to the data store
+    /// - Parameter handle: current `EventHandle` to process
+    private func handleLocationHintHandle(handle: EdgeEventHandle) {
+        guard let type = handle.type, EdgeConstants.JsonKeys.Response.EVENT_HANDLE_TYPE_LOCATION_HINT == type else { return }
+        guard let payload: [[String: Any]] = handle.payload else { return }
+
+        for locationHint in payload {
+            if let scope = locationHint[EdgeConstants.JsonKeys.Response.LocationHint.SCOPE] as? String, scope == "EdgeNetwork" {
+                guard let hint = locationHint[EdgeConstants.JsonKeys.Response.LocationHint.HINT] as? String, !hint.isEmpty else { continue }
```
addb, line 336: Do we have multiple hint and ttl values for EdgeNetwork scope? If not then should we break the loop instead of continue?

Sources/Edge.swift
```diff
@@ -238,4 +241,25 @@ public class Edge: NSObject, Extension {
...
+    /// Set the location hint for the Edge Network. The new location hint and expiry date (calculated from the ttlSeconds) are updated in memory and in
+    /// persistent storage. If the new location hint is different from the previous, then a shared state is also created with the new hint.
+    /// A nil `hint` value will clear the location hint in memory, persistent storage, and create a new shared state if a previous location hint was set.
+    /// - Parameters:
+    ///   - hint: the new EdgeNetwork location hint to set
+    ///   - ttlSeconds: the time-to-live for the location hint
+    private func setLocationHint(hint: String?, ttlSeconds: TimeInterval?) {
+        guard let state = state else { return }
+        if let hint = hint, let ttlSeconds = ttlSeconds {
+            state.setLocationHint(hint: hint, ttlSeconds: ttlSeconds, createSharedState: createSharedState(data:event:))
```
praveek, line 253: Do other extensions use the locationHint from the shared state?

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
```diff
@@ -152,7 +154,13 @@ class NetworkResponseHandler {
             if ignoreStorePayloads {
                 Log.debug(label: LOG_TAG, "Identities were reset recently, ignoring state:store payload for request with id: \(requestId)")
             } else {
-                handleStoreEventHandle(handle: eventHandle)
+                if let type = eventHandle.type {
+                    if EdgeConstants.JsonKeys.Response.EVENT_HANDLE_TYPE_STORE == type.lowercased() {
```
praveek, line 158: nit: Is `type.lowercased()` required? Calling it only for `if` condition is little confusing.

Sources/EdgeState.swift
```diff
@@ -74,4 +93,57 @@ class EdgeState {
         currentCollectConsent = status
```
praveek, line 93: This is not related to your PR but some of the variables are not guarded by DispatchQueue and this could cause race conditions and crashes.

praveek (approved):
+1

Sources/EdgeConstants.swift
kevinlind: Can you explain further? I guess I don't understand the question.

Sources/Edge.swift
kevinlind: Not currently and it's not expected as they should send requests through the Edge extension. The shared state is intended for Rules use cases where a customer may want to send a postback directly to the Edge Network. There's no hard requirement for that use-case either, however.

Sources/EdgeNetworkHandlers/NetworkResponseHandler.swift
kevinlind: I don't think it's required as the response uses "state:store" (lowercased). However, the original code used it so I preserved the logic.  The handle type "locationHint:result" is mix cased so I'm doing an exact match.

Sources/EdgeState.swift
kevinlind: I've updated EdgeState to guard the consent status. It's better to do it as it keep the design consistent, however I don't think the consent status is accessed from multiple threads. Can you take another look at this part of the PR?

Sources/EdgeConstants.swift
addb: It's fine. It is a nit point: Changing the location of constant.

Instead of Edge.LOCATION_HINT -> LocationHint.KEY

addb (approved):

PR #298, opened 2022-12-08
add update version script and gh action

sbenedicadb:


.github/workflows/update_versions.yml
```diff
@@ -0,0 +1,65 @@
+# This is a basic workflow that is manually triggered
+
+name: Update Versions
+
+# Controls when the action will run. Workflow runs when manually triggered using the UIk
```
emdobrin, line 5: typo: UI

.github/workflows/update_versions.yml
```diff
@@ -0,0 +1,65 @@
...
+# or API.
+on:
+  workflow_dispatch:
+    # Inputs the workflow accepts.
+    inputs:
+      version:
+        description: 'New version to use for the extension.  Example: 3.0.2'
```
emdobrin, line 12: nit: provide an example relevant for this extension, e.g 1.4.0

Script/update-versions.sh
```diff
@@ -0,0 +1,96 @@
...
+      ? ) help ;; # Print help in case parameter is non-existent
+   esac
+done
+
+# Print help in case parameters are empty
+if [ -z "$NAME" ] || [ -z "$NEW_VERSION" ]
+then
+   echo "********** USAGE ERROR **********"
+   echo "Some or all of the parameters are empty.  See usage below:";
+   help
```
emdobrin, line 50: you might want to exit early for missing required params when running this in a non-interactive mode like the gh action

.github/workflows/update_versions.yml
```diff
@@ -0,0 +1,65 @@
...
+        required: true
+        
+      branch:
+        description: 'Branch to be used when updating versions'
```
emdobrin, line 16: is there a way to include a branch picker here instead of text input?

PR #382, opened 2023-08-18
Update the validation expected results for recordSeparator in functional tests

cacheung:
MOB-19331
Update the validation expected results for recordSeparator. 
In iOS 17 and later, result displays " " (empty string). 
In iOS 16 and before, result displays "\u{0000}".

## Types of changes

## Checklist:


kevinlind (approved):
Looks good.
What you have is fine but another option would be to create a variable for the expected string then use the string in each test.
```
if #available(iOS 17, tvOS 17, *) {
   let expectedRecordSeparatorString = ""
} else {
   let expectedRecordSeparatorString = "\u{0000}"
}
```

cacheung:
> Looks good. What you have is fine but another option would be to create a variable for the expected string then use the string in each test.
> 
> ```
> if #available(iOS 17, tvOS 17, *) {
>    let expectedRecordSeparatorString = ""
> } else {
>    let expectedRecordSeparatorString = "\u{0000}"
> }
> ```

Thanks Kevin, I Iike the way you suggest.  Updated with that.

PR #413, opened 2023-09-28
Added upstream integration tests for configOverrides feature

addb:
## Description
Integration tests for configOverrides and Konductor responses.

Sample Integration test run: https://github.com/addb/aepsdk-edge-ios/actions/runs/6357026574/job/17267558942

Failing Tests: (Expecting 400 from Konductor, but for some reason getting 200)
- testSendEvent_withInvalidConfigOverrides_receivesExpectedNetworkResponseError
- testSendEvent_withInvalidConfigOverrides_notConfiguredValues_receivesExpectedNetworkResponseError
- testSendEvent_withInvalidConfigOverrides_dummyValues_receivesExpectedNetworkResponseError
- testSendEvent_withInvalidConfigOverrides_containingValidAndDummyValues_receivesExpectedNetworkResponseError

Enable above tests after PDCL-11131 issue is fixed:

Screenshot 2023-10-11 at 4 11 29 PM

## Types of changes

## Checklist:

### DIFF
--- a/Tests/TestUtils/TestBase.swift
+++ b/Tests/TestUtils/TestBase.swift
@@ -223,4 +223,25 @@ class TestBase: XCTestCase {
         guard !message.isEmpty && TestBase.debugEnabled else { return }
         print("TestBase - \(message)")
     }
+
+    // MARK: Specialized Helpers
+    func getEdgeEventHandles(expectedHandleType: String) -> [Event] {
+        return getDispatchedEventsWith(type: TestConstants.EventType.EDGE, source: expectedHandleType)
+    }
+
+    func getEdgeResponseErrors() -> [Event] {
+        return getDispatchedEventsWith(type: TestConstants.EventType.EDGE, source: TestConstants.EventSource.ERROR_RESPONSE_CONTENT)
+    }
+
+    /// Extracts the Edge location hint from the location hint result
+    func getLastLocationHintResultValue() -> String? {
+        let locationHintResultEvent = getEdgeEventHandles(expectedHandleType: TestConstants.EventSource.LOCATION_HINT_RESULT).last
+        guard let payload = locationHintResultEvent?.data?["payload"] as? [[String: Any]] else {
+            return nil
+        }
+        guard payload.indices.contains(2) else {
+            return nil
+        }
+        return payload[2]["hint"] as? String
+    }
 }
--- a/Tests/TestUtils/TestBase.swift
+++ b/Tests/TestUtils/TestBase.swift
@@ -223,25 +223,4 @@ class TestBase: XCTestCase {
         guard !message.isEmpty && TestBase.debugEnabled else { return }
         print("TestBase - \(message)")
     }
-
-    // MARK: Specialized Helpers
-    func getEdgeEventHandles(expectedHandleType: String) -> [Event] {
-        return getDispatchedEventsWith(type: TestConstants.EventType.EDGE, source: expectedHandleType)
-    }
-
-    func getEdgeResponseErrors() -> [Event] {
-        return getDispatchedEventsWith(type: TestConstants.EventType.EDGE, source: TestConstants.EventSource.ERROR_RESPONSE_CONTENT)
-    }
-
-    /// Extracts the Edge location hint from the location hint result
-    func getLastLocationHintResultValue() -> String? {
-        let locationHintResultEvent = getEdgeEventHandles(expectedHandleType: TestConstants.EventSource.LOCATION_HINT_RESULT).last
-        guard let payload = locationHintResultEvent?.data?["payload"] as? [[String: Any]] else {
-            return nil
-        }
-        guard payload.indices.contains(2) else {
-            return nil
-        }
-        return payload[2]["hint"] as? String
-    }
 }
--- a/Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
+++ b/Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
@@ -21,14 +21,9 @@ import XCTest
 class ConfigOverrideTests: TestBase {
     private var edgeEnvironment: EdgeEnvironment = getEdgeEnvironment()
     private var edgeLocationHint: EdgeLocationHint? = getLocationHint()
-
     private var networkService: RealNetworkService = RealNetworkService()
 
     let LOG_SOURCE = "ConfigOverrideTests"
-
-    let asyncTimeout: TimeInterval = 10
-
-    let url = "https://obumobile5.data.adobedc.net/ee/v1/interact"
 
     // Run before each test case
     override func setUp() {
@@ -52,12 +47,7 @@ class ConfigOverrideTests: TestBase {
         XCTAssertEqual(DispatchTimeoutResult.success, waitForRegistration.await(timeout: 2))
 
         // Set Edge location hint value if one is set for the test target
-        if edgeLocationHint != nil {
-            print("Setting Edge location hint to: \(String(describing: edgeLocationHint?.rawValue))")
-            Edge.setLocationHint(edgeLocationHint?.rawValue)
-        } else {
-            print("No preset Edge location hint is being used for this test.")
-        }
+        setInitialLocationHint(edgeLocationHint?.rawValue)
 
         resetTestExpectations()
         networkService.reset()
--- a/Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
+++ b/Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
@@ -42,8 +42,6 @@ class ConfigOverrideTests: TestBase {
         let waitForRegistration = CountDownLatch(1)
         MobileCore.setLogLevel(.trace)
 
-        print("Using Edge Network environment: \(edgeEnvironment)")
-
         // Set environment file ID for specific Edge Network environment
         MobileCore.configureWith(appId: getTagsEnvironmentFileId(for: edgeEnvironment))
 
--- a/Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
+++ b/Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
@@ -87,6 +87,7 @@ class ConfigOverrideTests: TestBase {
         XCTAssertEqual(200, matchingResponses.first?.responseCode)
     }
 
+    // TODO: Enable after PDCL-11131 issue is fixed
     // Test configOverrides with dummy data
     func testSendEvent_withInvalidConfigOverrides_receivesExpectedNetworkResponseError() {
         // Setup
@@ -122,6 +123,7 @@ class ConfigOverrideTests: TestBase {
         XCTAssertEqual(1, errorEvents.count)
     }
 
+    // TODO: Enable after PDCL-11131 issue is fixed
     // Tests ConfigOverrides with dataset not added in the datastream config and RSID not added to override setting in the Analytics upstream config
     func testSendEvent_withInvalidConfigOverrides_notConfiguredValues_receivesExpectedNetworkResponseError() {
         // Setup
@@ -167,6 +169,7 @@ class ConfigOverrideTests: TestBase {
         XCTAssertEqual(1, errorEvents.count)
     }
 
+    // TODO: Enable after PDCL-11131 issue is fixed
     func testSendEvent_withInvalidConfigOverrides_dummyValues_receivesExpectedNetworkResponseError() {
         // Setup
         let interactNetworkRequest = NetworkRequest(urlString: createInteractUrl(with: edgeLocationHint?.rawValue), httpMethod: .post)!
@@ -212,6 +215,7 @@ class ConfigOverrideTests: TestBase {
         XCTAssertEqual(1, errorEvents.count)
     }
 
+    // TODO: Enable after PDCL-11131 issue is fixed
     // test configOverrides with valid dataset ID, one valid and one dummy value for RSIDs
     func testSendEvent_withInvalidConfigOverrides_containingValidAndDummyValues_receivesExpectedNetworkResponseError() {
         // Setup
--- a/Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
+++ b/Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
@@ -19,8 +19,8 @@ import XCTest
 
 /// Performs validation on integration with the Edge Network upstream service
 class ConfigOverrideTests: TestBase {
-    private var edgeEnvironment: EdgeEnvironment = .prod
-    private var edgeLocationHint: EdgeLocationHint?
+    private var edgeEnvironment: EdgeEnvironment = getEdgeEnvironment()
+    private var edgeLocationHint: EdgeLocationHint? = getLocationHint()
 
     private var networkService: RealNetworkService = RealNetworkService()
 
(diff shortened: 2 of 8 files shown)


Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
```diff
@@ -569,6 +569,220 @@ class UpstreamIntegrationTests: TestBase {
...
     }
 
+    // Tests datastream config overrides
+    func testSendEvent_withValidConfigOverrides_receivesExpectedNetworkResponse() {
+        // Setup
+        // Note: test constructs should always be valid
+        let interactNetworkRequest = NetworkRequest(urlString: createURLWith(locationHint: edgeLocationHint), httpMethod: .post)!
+        // Setting expectation allows for both:
+        // 1. Validation that the network request was sent out
+        // 2. Waiting on a response for the specific network request (with timeout)
```
emdobrin, line 579: nit: let's remove these notes from tests, the intent is covered by using the corresponding framework APIs

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
```diff
@@ -569,6 +569,220 @@ class UpstreamIntegrationTests: TestBase {
...
+        networkService.setExpectationForNetworkRequest(networkRequest: interactNetworkRequest, expectedCount: 1)
+        let configOverrides = ["com_adobe_experience_platform": [
+                                    "datasets": [
+                                        "event": [
+                                            "datasetId": "do_not_delete_dataset2_e2e_testing_edge_konductor_configoverride"
```
emdobrin, line 584: this should be the datasetId and not the table name right?

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
```diff
@@ -569,6 +569,220 @@ class UpstreamIntegrationTests: TestBase {
...
+
+        // Verify
+        // Network response assertions
+        let matchingResponses = networkService.getResponsesFor(networkRequest: interactNetworkRequest, timeout: 5)
+
+        XCTAssertEqual(1, matchingResponses.count)
+        XCTAssertEqual(200, matchingResponses.first?.responseCode)
+    }
+
+    func testSendEvent_withInvalidConfigOverrides_invalidData_receivesExpectedNetworkResponseError() {
```
emdobrin, line 607: is the intention to test with both invalid config overrides and data or just testSendEvent_withInvalidConfigOverrides_receivesExpectedNetworkResponseError ?

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
```diff
@@ -569,6 +569,220 @@ class UpstreamIntegrationTests: TestBase {
...
+
+        // Test
+        Edge.sendEvent(experienceEvent: experienceEvent)
+
+        // Verify
+        // Network response assertions
+        let matchingResponses = networkService.getResponsesFor(networkRequest: interactNetworkRequest, timeout: 5)
+
+        XCTAssertEqual(1, matchingResponses.count)
+        XCTAssertEqual(400, matchingResponses.first?.responseCode)
```
emdobrin, line 626: For error responses can we also assert the error type and title, similar to testSendEvent_withInvalidDatastreamID_receivesExpectedError

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
```diff
@@ -569,6 +569,220 @@ class UpstreamIntegrationTests: TestBase {
...
+    }
+
+    func testSendEvent_withInvalidConfigOverrides_invalidDummyValues_receivesExpectedNetworkResponseError() {
```
emdobrin, line 666: Is this a duplicate of testSendEvent_withInvalidConfigOverrides_invalidConfigValues_receivesExpectedNetworkResponseError ?

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
```diff
@@ -569,6 +569,220 @@ class UpstreamIntegrationTests: TestBase {
...
+        // Setup
+        // Note: test constructs should always be valid
+        let interactNetworkRequest = NetworkRequest(urlString: createURLWith(locationHint: edgeLocationHint), httpMethod: .post)!
+        // Setting expectation allows for both:
+        // 1. Validation that the network request was sent out
+        // 2. Waiting on a response for the specific network request (with timeout)
+        networkService.setExpectationForNetworkRequest(networkRequest: interactNetworkRequest, expectedCount: 1)
+        // Overriding with dataset not added in the datastream config
+        // And Overriding with RSID not added to override setting in the Analytics upstream config
```
emdobrin, line 675: nit: I think you can move this as a description of the test for improved readability - same for all usages

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
```diff
@@ -569,6 +569,220 @@ class UpstreamIntegrationTests: TestBase {
...
+        // Verify
+        // Network response assertions
+        let matchingResponses = networkService.getResponsesFor(networkRequest: interactNetworkRequest, timeout: 5)
+
+        XCTAssertEqual(1, matchingResponses.count)
+        XCTAssertEqual(400, matchingResponses.first?.responseCode)
+    }
+
+    // test datastream ID override
+    func testSendEvent_withValidDatastreamID_receivesExpectedNetworkResponse() {
```
emdobrin, line 741: nit: 
..._withValidDatastreamIDOverride_..
..._withDummyDatastreamIdOverride_...

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
addb: Yes and the tests still passed

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
addb: I wanted to shwo the type of invalidConfigOverride, like using dummyValues. I see it might be confusing with the data parameter that we take in. I will update the name.

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
addb: in the invalidConfigValues I have an actual dataset and RSID but those are not added in the datastream config UI.

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
emdobrin: Is this a bug, are you tracking this down?

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
```diff
@@ -569,6 +569,267 @@ class UpstreamIntegrationTests: TestBase {
...
+        Edge.sendEvent(experienceEvent: experienceEvent)
+
+        // Verify
+        // Network response assertions
+        let matchingResponses = networkService.getResponsesFor(networkRequest: interactNetworkRequest, timeout: 5)
+
+        XCTAssertEqual(1, matchingResponses.count)
+        assertTypeMatch(expected: getAnyCodable(expectedErrorJSON)!,
+                        actual: getAnyCodable(matchingResponses.first?.responseString ?? ""),
+                        exactMatchPaths: ["status", "title", "type"])
```
emdobrin, line 635: 1. Looks like you are using exact match for all keys, you can simplify this to:
assertExactMatch(expected:..., actual: ...)

or you can include the other keys you may expect in the expected response and use type exact match just for status/title/type, and type match for all the others (see testSendEvent_withInvalidDatastreamID_receivesExpectedError as an example).
Same applies for the other tests below.

2. Is it possible to get more descriptive title/detail for these errors?

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
```diff
@@ -569,6 +569,267 @@ class UpstreamIntegrationTests: TestBase {
         assertExpectedEvents(ignoreUnexpectedEvents: true)
     }
 
+    // MARK: datastream config overrides test
```
emdobrin, line 572: Looks like you added a bunch of test cases for this feature, what do you think about moving them in a separate file ie ConfigOverride_UpstreamIntegrationTests?

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
addb: Hey @emdobrin, I am not getting errors yet m so I don't know what different errors we might see. I will update this once we get the error responses as expected.

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
addb: Yes, I was debating, should we have a new file, but it seems like a good idea since we have bunch of test cases.

Tests/TestUtils/TestBase.swift
```diff
@@ -223,4 +223,25 @@ class TestBase: XCTestCase {
...
+    func getLastLocationHintResultValue() -> String? {
+        let locationHintResultEvent = getEdgeEventHandles(expectedHandleType: TestConstants.EventSource.LOCATION_HINT_RESULT).last
+        guard let payload = locationHintResultEvent?.data?["payload"] as? [[String: Any]] else {
+            return nil
+        }
+        guard payload.indices.contains(2) else {
+            return nil
+        }
+        return payload[2]["hint"] as? String
+    }
```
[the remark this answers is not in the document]
addb: Since testBase is for Edge Network we could keep those edge specific still basic to be used across tests. I don't think it is a good idea to pass around testBase to helpers. 

If we decide to put testBase to a separate package, I think a better Idea would be to create a new helper which will extend the testBase class and add those helpers. 

```
// filename: TestBase+Helpers

extension TestBase {
 func getLastLocationHintResultValue() -> String? {
        let locationHintResultEvent = getEdgeEventHandles(expectedHandleType: TestConstants.EventSource.LOCATION_HINT_RESULT).last
        guard let payload = locationHintResultEvent?.data?["payload"] as? [[String: Any]] else {
            return nil
        }
        guard payload.indices.contains(2) else {
            return nil
        }
        return payload[2]["hint"] as? String
    }

}

File: Tests/UpstreamIntegrationTests/util/TestHelpers.swift
Comment (emdobrin) [inline Tests/UpstreamIntegrationTests/util/TestHelpers.swift:1]: renaming suggestion: TestSetupHelpers

File: Tests/UpstreamIntegrationTests/util/TestHelpers.swift [hunk 11]
```diff
@@ -0,0 +1,39 @@
...
+// governing permissions and limitations under the License.
+//
+
+import Foundation
+
+/// Creates a valid interact URL using the provided location hint.
+/// - Parameters:
+///    - locationHint: The location hint String to use in the URL
+/// - Returns: The interact URL with location hint applied
+public func createUrl(with locationHint: String?) -> String {
```
emdobrin, line 19: Since we moved this in a separate file I think it would make sense to call this createInteractUrl in case other helpers are added in the future

Tests/UpstreamIntegrationTests/UpstreamIntegrationTests.swift
```diff
@@ -570,61 +572,7 @@ class UpstreamIntegrationTests: TestBase {
...
-    ///    - locationHint: The location hint String to use in the URL
-    /// - Returns: The interact URL with location hint applied
-    private func createURLWith(locationHint: String?) -> String {
-        guard let locationHint = locationHint else {
-            return "https://obumobile5.data.adobedc.net/ee/v1/interact"
-        }
-        return "https://obumobile5.data.adobedc.net/ee/\(locationHint)/v1/interact"
-    }
-
     private func expectEdgeEventHandle(expectedHandleType: String, expectedCount: Int32 = 1) {
```
emdobrin, line 575: let's move this one too as Edge helper

Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
```diff
@@ -0,0 +1,327 @@
...
+@testable import AEPCore
+import AEPEdge
+import AEPEdgeIdentity
+import AEPServices
+import Foundation
+import XCTest
+
+/// Performs validation on integration with the Edge Network upstream service
+class ConfigOverrideTests: TestBase {
+    private var edgeEnvironment: EdgeEnvironment = .prod
```
emdobrin, line 22: We should have a constant for the DEFAULT_EDGE_ENVIRONMENT and use for all tests

Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
```diff
@@ -0,0 +1,327 @@
...
+        })
+        XCTAssertEqual(DispatchTimeoutResult.success, waitForRegistration.await(timeout: 2))
+
+        // Set Edge location hint value if one is set for the test target
+        if edgeLocationHint != nil {
+            print("Setting Edge location hint to: \(String(describing: edgeLocationHint?.rawValue))")
+            Edge.setLocationHint(edgeLocationHint?.rawValue)
+        } else {
+            print("No preset Edge location hint is being used for this test.")
+        }
```
emdobrin, line 66: this can be an EdgeHelper too

Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
```diff
@@ -0,0 +1,327 @@
...
+    override func setUp() {
+        ServiceProvider.shared.networkService = networkService
+
+        super.setUp()
+
+        continueAfterFailure = true
+        TestBase.debugEnabled = true
+        // Extract Edge Network environment level from shell environment; see init for default value
+        self.edgeEnvironment = EdgeEnvironment()
+        print("Using Edge Network environment: \(edgeEnvironment.rawValue)")
```
emdobrin, line 43: this can be a helper too and it can return the shell variable or the default environment

Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
addb: What are you thinking when you say as a helper? I have pushed some updates, but I think I might have not extracted this lcoationHint and edgeEnvironment as per your thinking. Let me know if I missed something in the updates.

Tests/UpstreamIntegrationTests/util/TestSetupHelpers.swift
```diff
@@ -0,0 +1,53 @@
...
+    print("Using Edge Network environment: \(edgeEnvironment.rawValue)")
+
+    return edgeEnvironment
+}
+
+/// Creates a valid interact URL using the provided location hint.
+/// - Parameters:
+///    - locationHint: The location hint String to use in the URL
+/// - Returns: The interact URL with location hint applied
+public func createInteractUrl(with locationHint: String?) -> String {
```
emdobrin, line 33: Since these are test setup helpers based on the input params, I think this one should go to TestBase+EdgeHelpers

Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
emdobrin: I was referring to L56 -> L62 which can be extracted in TestBase+EdgeHelpers in a simple func setInitialLocationHint(edgeLocationHint) so all tests can use it as is and also save some lines

Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
```diff
@@ -0,0 +1,323 @@
...
+
+        super.setUp()
+
+        continueAfterFailure = true
+        TestBase.debugEnabled = true
+
+        let waitForRegistration = CountDownLatch(1)
+        MobileCore.setLogLevel(.trace)
+
+        print("Using Edge Network environment: \(edgeEnvironment)")
```
emdobrin, line 45: nit: can be removed from all tests, you already print this line inside getEdgeEnvironment

emdobrin (approved):
Left some responses below, looks great

Tests/UpstreamIntegrationTests/ConfigOverrideTests.swift
addb: I did not move that since we don't register Edge extension in the TestBase.

Tests/UpstreamIntegrationTests/util/TestSetupHelpers.swift
addb: I kept this here since it doesn't deal with events (Extension related) and is purely string helper.

addb: But I see your point. I will move it.

PR #458, opened 2024-04-17
Cache configuration with Edge hit (resolves #367)

### LINKED ISSUE #367: Cache config with Edge event
### Task description

For consistency with Android, we should cache the Edge config settings with the Event. 
- Update EventDataEntity to include config and update EdgeHitProcessor to use this instead of retrieving right before processing.
- Update relevant tests and/or add new ones.

### Additional implementation details or code snippet(s)

_No response_

kevinlind:
## Description

Cache the Edge configuration with the queued data entity instead of retrieving the configuration when the hit is processed. This update preserves the previous EdgeHitProcessor workflow to handle cases where queued data entities without a configuration are present after upgrading to this newer version.

- Edge extension will now bundle the edge configuration in the EdgeDataEntity when adding entity to queue.
- EdgeHitProcessor now reads the edge configuration from the EdgeDataEntity instead of getting it from the current Configuration shared state.
- EdgeHitProcess continues to use old workflow of validating against `readyForEvent` and reading the Configuration shared state if the current data entity contains an empty configuration.

## Types of changes

## Checklist:

### DIFF
--- a/Sources/Edge.swift
+++ b/Sources/Edge.swift
@@ -129,7 +129,7 @@ public class Edge: NSObject, Extension {
                 $0.key == EdgeConstants.SharedState.Configuration.EDGE_DOMAIN
         } as [String: Any]
 
-        if edgeConfig[EdgeConstants.SharedState.Configuration.CONFIG_ID] == nil || (edgeConfig[EdgeConstants.SharedState.Configuration.CONFIG_ID] as? String ?? "").isEmpty {
+        if (edgeConfig[EdgeConstants.SharedState.Configuration.CONFIG_ID] as? String ?? "").isEmpty {
             Log.warning(label: EdgeConstants.LOG_TAG, "\(SELF_TAG) - Unable to process the event '\(event.id.uuidString)', Configuration is missing edge.configId.")
             return // drop current event
         }
--- a/Sources/Edge.swift
+++ b/Sources/Edge.swift
@@ -237,7 +237,6 @@ public class Edge: NSObject, Extension {
         let hitProcessor = EdgeHitProcessor(networkService: networkService,
                                             networkResponseHandler: networkResponseHandler,
                                             getSharedState: getSharedState(extensionName:event:),
-                                            getXDMSharedState: getXDMSharedState(extensionName:event:barrier:),
                                             readyForEvent: readyForEvent(_:),
                                             getImplementationDetails: getImplementationDetails,
                                             getLocationHint: getLocationHint,
--- a/Sources/Edge.swift
+++ b/Sources/Edge.swift
@@ -96,27 +96,21 @@ public class Edge: NSObject, Extension {
     /// - Parameter event: an event containing ExperienceEvent data for processing
     func handleExperienceEventRequest(_ event: Event) {
         guard !shouldIgnore(event: event) else { return }
-
-        guard let data = event.data, !data.isEmpty else {
-            Log.trace(label: EdgeConstants.LOG_TAG, "\(SELF_TAG) - Event with id \(event.id.uuidString) contained no data, ignoring.")
-            return
-        }
-
         processAndQueueEvent(event: event)
     }
 
     /// Handles the Consent Update event
     /// - Parameter event: current event to process
     func handleConsentUpdate(_ event: Event) {
-        guard let data = event.data, !data.isEmpty else {
-            Log.trace(label: EdgeConstants.LOG_TAG, "\(SELF_TAG) - Consent update request event \(event.id.uuidString) contained no data, ignoring.")
-            return
-        }
-
         processAndQueueEvent(event: event)
     }
 
     private func processAndQueueEvent(event: Event) {
+        guard let data = event.data, !data.isEmpty else {
+            Log.trace(label: EdgeConstants.LOG_TAG, "\(SELF_TAG) - Event with id \(event.id.uuidString) contains no data, ignoring.")
+            return
+        }
+
         // get Configuration shared state, this should be resolved based on readyForEvent check
         guard let edgeConfig = getEdgeConfig(event: event) else {
             Log.warning(label: EdgeConstants.LOG_TAG,
--- a/Sources/Edge.swift
+++ b/Sources/Edge.swift
@@ -117,20 +117,10 @@ public class Edge: NSObject, Extension {
     }
 
     private func processAndQueueEvent(event: Event) {
-        // get configuration needed for Edge hit, this should be resolved based on readForEvent check
-        guard let configurationState = getSharedState(extensionName: EdgeConstants.SharedState.Configuration.STATE_OWNER_NAME, event: event)?.value else {
-            Log.warning(label: EdgeConstants.LOG_TAG, "\(SELF_TAG) - Unable to process the event '\(event.id.uuidString)', Configuration shared state is nil.")
-            return // drop current event
-        }
-
-        let edgeConfig = configurationState.filter {
-            return $0.key == EdgeConstants.SharedState.Configuration.CONFIG_ID ||
-                $0.key == EdgeConstants.SharedState.Configuration.EDGE_ENVIRONMENT ||
-                $0.key == EdgeConstants.SharedState.Configuration.EDGE_DOMAIN
-        } as [String: Any]
-
-        if (edgeConfig[EdgeConstants.SharedState.Configuration.CONFIG_ID] as? String ?? "").isEmpty {
-            Log.warning(label: EdgeConstants.LOG_TAG, "\(SELF_TAG) - Unable to process the event '\(event.id.uuidString)', Configuration is missing edge.configId.")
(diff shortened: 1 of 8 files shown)


Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -61,10 +61,22 @@ class EdgeHitProcessor: HitProcessing {
...
 
         let event = edgeEntity.event
-        guard readyForEvent(event) else {
-            Log.debug(label: EdgeConstants.LOG_TAG, "\(SELF_TAG) - Not ready for event, will retry hit with id '\(entity.uniqueIdentifier)'.")
-            completion(false)
-            return
+        var edgeConfig: [String: String]
+
+        // Check which workflow is used to obtain configuration
+        if edgeEntity.configuration.isEmpty {
```
emdobrin, line 67: recommend reversing the order of this condition and handling the most common case first, where the entity has config, and the old workflow after (rare post upgrade)

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
```diff
@@ -94,10 +106,14 @@ class EdgeHitProcessor: HitProcessing {
...
+    ///   - eventConfig: configuration data for this event.
     ///   - requestBuilder: the `RequestBuilder` object to build the request payload.
     ///   - completion: completion handler to notify the caller about the hit response.
-    private func processExperienceEvent(entityId: String, event: Event, requestBuilder: RequestBuilder, completion: @escaping (Bool) -> Void) {
-        guard let edgeConfig = getEdgeConfig(event: event), var datastreamId = edgeConfig[EdgeConstants.SharedState.Configuration.CONFIG_ID] else {
+    private func processExperienceEvent(entityId: String, event: Event, edgeConfig: [String: String], requestBuilder: RequestBuilder, completion: @escaping (Bool) -> Void) {
+        guard var datastreamId = edgeConfig[EdgeConstants.SharedState.Configuration.CONFIG_ID], !datastreamId.isEmpty else {
+            Log.warning(label: EdgeConstants.LOG_TAG,
+                        "\(SELF_TAG) - Unable to process the event '\(event.id.uuidString)' " +
+                            "due to invalid edge.configId in configuration.")
```
emdobrin, line 116: - rewording - let's clarify what is the exact invalid scenario here: due to invalid ... -> due to missing/empty ...
- I assume this will only be possible for the old path since this verification can be done prior to queueing. could we 1. ensure queued config is valid 2. do this check for old workflow and finalize the config validity even before calling these methods

Sources/Edge.swift
```diff
@@ -102,6 +102,38 @@ public class Edge: NSObject, Extension {
...
+        let edgeConfig = configurationState.filter {
+            return $0.key == EdgeConstants.SharedState.Configuration.CONFIG_ID ||
+                $0.key == EdgeConstants.SharedState.Configuration.EDGE_ENVIRONMENT ||
+                $0.key == EdgeConstants.SharedState.Configuration.EDGE_DOMAIN
+        } as [String: Any]
+
+        if (edgeConfig[EdgeConstants.SharedState.Configuration.CONFIG_ID] as? String ?? "").isEmpty {
+            Log.warning(label: EdgeConstants.LOG_TAG, "\(SELF_TAG) - Unable to process the event '\(event.id.uuidString)', Configuration is missing edge.configId.")
+            return // drop current event
+        }
```
emdobrin, line 135: recommend extracting in a separate func getConfig, if null return.

Sources/Edge.swift
```diff
@@ -102,6 +102,38 @@ public class Edge: NSObject, Extension {
...
+        processAndQueueEvent(event: event)
+    }
+
+    /// Handles the Consent Update event
+    /// - Parameter event: current event to process
+    func handleConsentUpdate(_ event: Event) {
+        guard let data = event.data, !data.isEmpty else {
+            Log.trace(label: EdgeConstants.LOG_TAG, "\(SELF_TAG) - Consent update request event \(event.id.uuidString) contained no data, ignoring.")
+            return
+        }
```
emdobrin, line 114: could be moved in processAndQueueEvent and reduce few lines

Sources/EdgeNetworkHandlers/EdgeHitProcessor.swift
kevinlind: The check on configId is only needed for the old path, as the configId is validated to be non-nil/non-empty before queuing the data entity. However, I still need to extract the configId in processExperienceEvent/processConsentEvents so it's natural to use a `guard` for this and simple enough to include the `!isEmpty` check. If I were to perform the check earlier, I'd still need to unwrap the `configId` from the EdgeDataEntity, so I'd still need to include a `guard` or some other such statement.

addb (approved):

Tests/UnitTests/EdgeHitProcessorTests.swift
```diff
@@ -87,131 +83,76 @@ class EdgeHitProcessorTests: XCTestCase, AnyCodableAsserts {
...
+    // Mock for `getXDMSharedState` closure in `EdgeHitProcessor`. Reassign variable in test to change behavior.
+    private var mockGetXDMSharedState: (String, Event?, Bool) -> SharedStateResult? = {extensionName, _, _ in
+        XCTFail("Test called 'getXDMSharedState(\(extensionName))' but was not expected.")
         return nil
     }
 
-    private func readyForEvent(_ event: Event) -> Bool {
-        return true
+    // Mock for `readyForEvent` closure in `EdgeHitProcessor`. Reassign variable in test to change behavior.
+    private var mockReadForEvent: (Event) -> Bool = {_ in
```
emdobrin, line 102: typo: mockReadyForEvent

Tests/UnitTests/EdgeExtensionTests.swift
```diff
@@ -218,13 +327,79 @@ class EdgeExtensionTests: XCTestCase, AnyCodableAsserts {
...
+        guard let data = dataEntity.data, let edgeDataEntity = try? JSONDecoder().decode(EdgeDataEntity.self, from: data) else {
+            XCTFail("Unable to decode EdgeEntity from DataEntity!")
+            return
+        }
+
+        // Verify EdgeDataEntity only contains config ID, domain, and environment
+        XCTAssertEqual(3, edgeDataEntity.configuration.count)
+        XCTAssertEqual("12345-example", edgeDataEntity.configuration[EdgeConstants.SharedState.Configuration.CONFIG_ID])
+        XCTAssertEqual("edge.com", edgeDataEntity.configuration[EdgeConstants.SharedState.Configuration.EDGE_DOMAIN])
+        XCTAssertEqual("dev", edgeDataEntity.configuration[EdgeConstants.SharedState.Configuration.EDGE_ENVIRONMENT])
```
emdobrin, line 369: for improved readability, can we use the json asserts instead for these tests

Tests/FunctionalTests/EdgeQueuedEntityFunctionalTests.swift
```diff
@@ -0,0 +1,156 @@
...
+
+        // Setup network response and expected network requests
+        let responseConnection: HttpConnection = HttpConnection(data: "{\"test\": \"json\"}".data(using: .utf8),
+                                                                response: HTTPURLResponse(url: exEdgeInteractProdUrl,
+                                                                                          statusCode: 200,
+                                                                                          httpVersion: nil,
+                                                                                          headerFields: nil),
+                                                                error: nil)
+        mockNetworkService.setMockResponse(url: TestConstants.EX_EDGE_INTERACT_PROD_URL_STR, httpMethod: HttpMethod.post, responseConnection: responseConnection)
+        mockNetworkService.setExpectationForNetworkRequest(url: TestConstants.EX_EDGE_INTERACT_PROD_URL_STR, httpMethod: HttpMethod.post, expectedCount: 1)
```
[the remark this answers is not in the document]
emdobrin: covered in unit tests

Sources/EdgeNetworkHandlers/SharedStateReader.swift
```diff
@@ -0,0 +1,53 @@
...
+    private let SELF_TAG = "SharedStateReader"
+    private var getSharedState: (String, Event?, Bool) -> SharedStateResult?
+
+    init(getSharedState: @escaping (String, Event?, Bool) -> SharedStateResult?) {
+        self.getSharedState = getSharedState
+    }
+
+    /// Get the Edge configuration by quering the Configuration shared state and filtering out only the key needed for Edge requests.
+    /// - Parameter event: the `Event` to get the configuration
+    /// - Returns: A dictionary of Edge configuration values, or nil if the Configuration shared state could not be retrieved or if the `edge.configId` key is missing or empty.
```
emdobrin, line 26: doesn't seem to return nil 🤓

emdobrin (approved):
minor comment on docs, looks good

